PR #26, opened 2022-04-01
CSS cleanup and misc. fixes

astrotars:
@elof There was CSS overriding the default top nav functionality. I cleaned that up and a few other styles throughout. In a mobile viewport, the nav items end up in the hamburger slide out but they're nested a level back (default behavior and the only way around this is a custom theme). 

Screen Shot 2022-03-31 at 8 21 01 PM

Screen Shot 2022-03-31 at 8 20 48 PM

Screen Shot 2022-03-31 at 8 20 39 PM

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -60,7 +60,6 @@ const config = {
         searchParameters: {},
       },
       navbar: {
-
         logo: {
           alt: 'Macrometa Logo',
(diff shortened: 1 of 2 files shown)


astrotars:
@elof Customizing Stoplight Elements will take a bit more time. It seems... tricky...

elof (approved):
:shipit:

PR #34, opened 2022-04-12
Add OG image and update title/description

astrotars:


astrotars:
@elof what do you think about getting a custom image designed for the docs open-graph image? It could be as simple as our logo + "docs" and a fun illustration behind it. I'm hoping that our docs become something developers want to share with the world. For now, this PR adds a title, description, and image pulled from the website. Image below for reference.

map

elof:
I'm definitely down for some fun design for the docs. @dustinlarimer something to noodle on here :)

elof (approved):
:shipit:

PR #37, opened 2022-04-12
Enable Algolia Docsearch

astrotars:
@elof this will enable the search and run a new crawl. It does not account for REST API search yet (this one is complicated but I'm looking into it).


elof (approved):
:shipit:

PR #62, opened 2022-05-24
Create CODEOWNERS

jamesm-macrometa:
Created a CODEOWNERS file for the DOCS repository.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -3,7 +3,7 @@
 
 # These owners will be the default owners for everything in
 # the docs folder. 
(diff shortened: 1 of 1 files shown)


diana-macrometa (approved):
Looks good. We should test it with some minor PRs, and I might take myself off as owner of everything at some point, but let's see where this gets us.

jamesm-macrometa:
> Looks good. We should test it with some minor PRs, and I might take myself off as owner of everything at some point, but let's see where this gets us.

FYI I already removed you as a top-level owner and added @astrotars instead.  You and I are only owners of the /docs/ folder.

PR #75, opened 2022-06-08
Update spec.json

Luba89:
PR regarding https://macrometa.atlassian.net/browse/DOC-289


static/swagger/spec.json
```diff
@@ -14155,7 +14155,7 @@
...
...*padded:** generates keys of a fixed length (16 bytes) in ascending lexicographical sort order.\n      - **uuid64**: generates universally unique keys.\n    - **increment**: increment value for *autoincrement* key generator. Not used for other key generator types.\n    - **offset**: Initial offset value for *autoincrement* key generator. Not used for other key generator types.\n  - **isSystem**: If *true*, create a system collection. The *collection-name* must start with an underscore. Not Applicable for *end-users*. (The default is *false*)\n  - **isLocal**: If *true*, create a local collection. For a local collection data is not replicated across regions. (The default is *false*)\n - **type**: The type of the collection to be created.\n The following values for *type* are valid (The default is *2*):\n    - *2*: document collection\n    - *3*: edge collection.\n  - **stream**: If *true*, create a local stream for collection. (The default is *false*)\n  - **shardKeys**: The specified shard key determines in which shard a given document is to be stored. Choosing the right shard key can have significant impact on your performance can reduce network traffic and increase performance.",
```
jamesm-macrometa:
```suggestion
        "description": "Create collection.\nRequires a JSON object with these properties:\n- **name**: The name of the collection.\n - **keyOptions**:\n    - **allowUserKeys**: If set to *true*, you can supply a key value in the *_key* attribute of a document. If set to *false*, the key generator creates keys, and an error occurs if you add a *_key* value.\n    - **type**: Choose the type of key generator:\n      - **traditional**: Generates numerical keys in ascending order.\n      - **autoincrement:** Generates numerical keys in ascending order with configurable initial offset and spacing.\n      - **padded:** Generates keys of a fixed length (16 bytes) in ascending lexicographical sort order.\n      - **uuid64**: Generates universally unique keys.\n    - **increment**: Increment value for *autoincrement* key generator. Not used for other key generator types.\n    - **offset**: Initial offset value for *autoincrement* key generator. Not used for other key generator types.\n  - **isSystem**: If *true*, create a system collection. The *collection-name* must start with an underscore. Not Applicable for *end-users*. (The default is *false*)\n  - **isLocal**: If *true*, create a local collection. For a local collection data is not replicated across regions. (The default is *false*)\n - **type**: The type of the collection to be created.\n The following values for *type* are valid (The default is *2*):\n    - *2*: document collection\n    - *3*: edge collection.\n  - **stream**: If *true*, create a local stream for collection. (The default is *false*)\n  - **shardKeys**: The specified shard key determines in which shard a given document is to be stored. Choosing the right shard key can have significant impact on your performance can reduce network traffic and increase performance.",
```

jamesm-macrometa: Only change here is adding initial capitalization to a few sentences.  Looks good!

jamesm-macrometa (changes requested):
One minor change, otherwise looks good to me.

static/swagger/spec.json
jamesm-macrometa: Oh, just noticed this is the Docs repo and not a Swagger repo!  Incorporating changes and resolving PR.

jamesm-macrometa (approved):

astrotars (approved):

PR #77, opened 2022-06-08
DOC-278 - Move appendix content

diana-macrometa:
Reference content was buried in an Appendix folder in the Documents collection docs. It did not belong there.

This PR does the following: 
- Moves the content to the top level and renames the folder to References
- Moves Naming Conventions to Essentials
- Removes capitalization from a couple filenames
- No content changes


jamesm-macrometa (approved):
Looks good to me!

PR #81, opened 2022-06-09
Fix Introduction

jamesm-macrometa:
Introduction example box used old Example box style.  Also fixed position on sidebar.


diana-macrometa (approved):
LGTM

PR #100, opened 2022-06-17
Split stream worker functions into multiple sections

jamesm-macrometa:
The purpose of this PR is to split the Stream Worker Functions section into multiple sections to make it easier to navigate and scan.


jamesm-macrometa:
@diana-macrometa You can review this if you want, but not necessary as the content is huge and we will be creating a task to review it all in-depth anyways.

diana-macrometa (approved):
Epic work, James!

PR #120, opened 2022-07-11
change case of file names

elof:
This PR changes the case for folders and files that were capitalized to lower case

### DIFF
--- a/docs/essentials/authentication.md
+++ b/docs/essentials/authentication.md
@@ -1,0 +1,105 @@
+---
+sidebar_position: 2
+title: Authentication
+slug: authentication
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+There are multiple ways users can access their Macrometa GDN account:
+
+* User Authentication
+* Token based Authentication
+* API Keys
+
+## User Authentication
+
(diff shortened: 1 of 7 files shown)


diana-macrometa (approved):
Thank you! I was just going to change them as we touched them. :)

diana-macrometa:
Any idea why the check is failing?

elof:
It looks like docusarus is recognizing both Capital.md and capital.md as the same route. Looking into it

diana-macrometa:
Yeah, it does that. I usually make a change to the tense at the same time as I change the content. That seems to goose it into actually doing the update.

Also, GitHub doesn't recognize changing the file capitalization as a change, IIRC. Makes it even more annoying.

elof:
I think Nick ran into this same problem and found out that it was because github actions keeps a cache of the file name route paths. I think we might need to wait until the cache clears, because if I remember correctly there isn't a way to force github actions cache to clear

elof:
I did some research. If this is related to cache then it will reset in 7 days. There is a way to reset the cache but it requires admin rights. Since this isn't critical I suggest we let it be for now and try to run the build again next week

elof:
closing this as the other PR fixed the cases

PR #131, opened 2022-07-15
add api reference to create doc collection page

elof:
I think I came up with a low-effort way to get it in there. We can link to the API reference. It has code generation in 18 programming languages and is executable. It will also be automatically updated via our swagger file. I also reorganized the page to focus on the API and SDKs (renamed the hotlink from code to SDK) first.


diana-macrometa:
Two issues here:
- Generally organization goes from simplest to hardest. If we are aiming to speak to users who are noobie developer level, will this be starting with the easiest version of the task for them?
- If we do this here, then to be consistent, we should also do the same thing for the other three types of collection.

elof:
@diana-macrometa I generally see docs focused on using the API first. If you look at Stripe (the conical example everyone loves to reference), their main focus is on using the API in all their docs. Ideally, the console is intuitive enough not to need documentation. We aren't quite there yet, but we will be!

elof:
@diana-macrometa I think we can use this pattern for all of the docs

diana-macrometa:
Actually, Stripe docs, at least in the random example I accessed, seem to put the dashboard and API at the same level, just different tabs.

Is there any way we could use the tab plugin to do this? It's a really great solution. I still think we should have the console/UI be first, but then we could have ALL of the options be right at the top after the introductory paragraphs.

image

elof:
@diana-macrometa ah, I guess I was looking at a part of the docs that didn't include. We could update the SDK tabs to say Dashboard/JS SDK/Python SDK instead of JS/Python

diana-macrometa:
@elof, THAT is a pattern I can get behind! Bonus points, it is easy to turn into a template!

Would it make sense to add Query (C8QL), API, and CLI tabs as well? I am pretty sure that most simple interactions with collections can be done with those as well. Maybe complex ones too, IDK.

elof:
@diana-macrometa not sure if query would make sense. We could use the API reference to generate cURL examples, or we could still just link to the API reference were there are a bunch of code gen options. We could definitely add CLI to the tabs as well

elof:
@diana-macrometa what do you think of this: 

Screen Shot 2022-07-15 at 1 16 53 PM

diana-macrometa:
Needs some text edits, but the arrangement is good. Let's do it!

@dustinlarimer and @jamesm-macrometa and @james-macrometa , what do you think?

docs/collections/documents/create-document-store.md
```diff
@@ -8,8 +8,16 @@ import TabItem from '@theme/TabItem';
 
 This page explains how to create a new Document Store collection.
 
-## Create a Document Store Collection with the Console
+Use our interactive API Reference with code generation in 18 programming languages to [create a Document Store Collection](https://macrometa.com/docs/api#/operations/handleCommandPost:CreateCollection). 
 
+### Create a Document Store Collection with an SDK or in the web console.
```
diana-macrometa:
```suggestion
### Create a Document Store Collection
```

docs/collections/documents/create-document-store.md
```diff
@@ -8,8 +8,16 @@ import TabItem from '@theme/TabItem';
...
+
+The below example shows the steps for connecting a fabric and then creating a collection called `employees`.
```
diana-macrometa:
```suggestion
```

docs/collections/documents/create-document-store.md
```diff
@@ -39,7 +44,7 @@ The below example shows the steps for connecting a fabric and then creating a co
 ```
 
 </TabItem>
-<TabItem value="js" label="Javascript">
+<TabItem value="js" label="Javascript SDK">
```
diana-macrometa, line 45: The below example shows the steps for connecting a fabric and then creating a collection called `employees`.

docs/collections/documents/create-document-store.md
```diff
@@ -23,13 +31,10 @@ Follow these instructions to create a new collection using the GDN console web U
...
-
-The below example shows the steps for connecting a fabric and then creating a collection called `employees`.
-
-<Tabs groupId="operating-systems">
-<TabItem value="py" label="Python">
+```
 
+</TabItem>
+<TabItem value="py" label="Python SDK">
 ```py
```
diana-macrometa, line 35:
```suggestion
The below example shows the steps for connecting a fabric and then creating a collection called `employees`.

```py
```

diana-macrometa:
FYI, the two code examples were taken from a tutorial. I have a Jira ticket to make them generic and add a bit more explanation, but figured having something was better than nothing.

elof:
@diana-macrometa added one little fix to the py tab

jamesm-macrometa (approved):

jamesm-macrometa:
@diana-macrometa Are we ready to merge this?

diana-macrometa:
Once the conflicts are resolved, yes.

elof:
I'll work on resolving the conflicts when I get back from vacation

jamesm-macrometa:
@elof What exactly is the conflict here? Do we just need to rename/move the listed file:

`docs/collections/documents/create-document-store.md`

If that's all there is too it, I can do it myself. But if it's more complicated I can leave you to it. Let me know

cc @diana-macrometa 

Edit - Never mind, just noticed I can view the exact conflict through GitHub.  This looks fairly straight forward - want me to handle it Justin?

elof:
@jamesm-macrometa I accidentally changed the same file in a different PR, so it broke this one :/

PR #138, opened 2022-07-18
Fixing Tab view not available on streamworker example page

koshy-thomas:
Tab view for code blocks not available in the following page:
https://macrometa.com/docs/cep/stream-workers-example

Fixing the same, turns out there was a small missing part of importing here. Also there was a ending backtick missing in the last piece of js code snippet.

### DIFF
--- a/docs/cep/stream-workers-example.md
+++ b/docs/cep/stream-workers-example.md
@@ -3,6 +3,8 @@ sidebar_position: 10
 title: Stream Workers Example
 ---
 
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
 
 Assume the following credentials:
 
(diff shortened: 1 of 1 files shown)


docs/cep/stream-workers-example.md
```diff
@@ -629,7 +631,7 @@ The following example uses the code snippets provided in this tutorial.
             @info(name='Dump')
             INSERT INTO SampleCargoAppDestTable
             SELECT weight
-            FROM SampleCargoAppInputTable;
+            FROM SampleCargoAppInputTable;`
```
diana-macrometa, line 634: Is the addition here intentional? It does not look correct.

koshy-thomas: Yes it's intentional

koshy-thomas: There was a missing backtick here, the ending backtick for line 583/585 is missing
const updatedAppDefinition = `

The ending backtick should be there in line 634 as added

dustinlarimer: Nice catch @koshy-thomas :+1:

dustinlarimer (approved):

dustinlarimer:
Existing page: https://macrometa.com/docs/cep/stream-workers-example

Fixed by this PR: https://docs-git-koshymm-team-macrometa.vercel.app/cep/stream-workers-example

PR #140, opened 2022-07-19
Free Tier Views per Fabric update

james-macrometa:
Search is not a feature available to the free tier. I have make a correction to reflect that information.


diana-macrometa (approved):
Good catch!

PR #144, opened 2022-07-21
Update release-notes-0-17-10.md

diana-macrometa:
Add Kafka to release notes.

Reviewers - Kafka is already mentioned in Stream Workers, which suggests it was already there. Please update the release notes description to help the user understand the scope of the change and what they will be able to do now that they couldn't before.


anandani-manish:
Added @chandrashekhar-macrometa 
Hi @chandrashekhar-macrometa  we need to work with @diana-macrometa to identify a user facing tutorial for Kafka on Pulsar

diana-macrometa:
Reviewers, is this release note entry good? It seems like there could be more detail added. What would be useful for the user to know about this change?

PR #180, opened 2022-08-10
pyc8 changes for bulkupdaterestql kv

koshy-thomas:
There needs to be some wait time needed when creating/updating a RestQL before executing it. Hence, I have added time.sleep wherever necessary also added example implementation of update_restql and delete_restql methods to cover all the restql methods provided by the pyC8 driver.


jamesm-macrometa (approved):
Looks good!

PR #181, opened 2022-08-10
pyc8 streamworker restapi audit

koshy-thomas:
The create_connection method to connect to the websocket url needed additional authorization headers to be passed as arguments:
create_connection(producerurl,header={"content-type": 'application/json', 'authorization': AUTH_TOKEN})

The time.sleep(20)  which makes it wait for 20secs for the changes to initialize is too much so I have changed it to 5 secs (time.sleep(5)) 
Also changed all .format() string to f-strings.


jamesm-macrometa (approved):
LGTM!

PR #188, opened 2022-08-10
Revert "Change top-level sidebar positions to double digits"

jamesm-macrometa:
Reverts Macrometacorp/docs#186


jamesm-macrometa:
I noticed this change screwed up the overall order of the sidebar.  Reverting this change.

PR #191, opened 2022-08-11
Update return-data-in-batches.md

bdzekic1:
Since pyC8 code part is not past practice due to issues with the SDK with this example, it was suggested that it should be removed for the time being, until the issue has been resolved.


diana-macrometa:
Please add a note to the document explaining the change. Format is:
```
:::note
NOTE TEXT
:::
```

jamesm-macrometa (approved):
LGTM!

PR #196, opened 2022-08-12
Refactoring constants to upper case

edgargarciamacrometa:
Also, I added a timeout for the client.on_change() method and changed the .format to f"string"


docs/collections/documents/tutorials/using-realtime-updates.md
```diff
@@ -67,73 +67,72 @@ Let's assume your
...
-      {"ip": "40.1.1.4", "action": "block", "rule": "blacklistA"},
-      {"ip": "50.1.1.5", "action": "block", "rule": "blacklistB"},
-    ]
+import time
+import threading
+import pprint
+from c8 import C8Client
+
+# Variables - URLs
+GLOBAL_URL = "gdn.paas.macrometa.io"
```
diana-macrometa, line 76: Why are you making the variables uppercase? I thought lowercase was the accepted standard.

edgargarciamacrometa: According to [PEP](https://peps.python.org/pep-0008/#constants), constants should be written in uppercase. The support team has been making changes for the pyC8 audit using as a guide the previous link to keep an internal standard. There is another [file](https://github.com/Macrometacorp/docs/blob/master/docs/streams/tutorials/using-realtime-updates.md) within the docs, which has the same content as this. Maybe part of the concern is that we have a #Variables comment in the python file, but they are constants since they do not change their value over time for this example.

diana-macrometa: Cool. I didn't realize variables needed to change value in order to be variables.  Learn something new every day. :)

Would they not be variables because the constant would change for different users? Like my email is different than your email, so our variable/constant would have a different value.

diana-macrometa (approved):

docs/collections/documents/tutorials/using-realtime-updates.md
edgargarciamacrometa: I understand your point, and I think it can even get more complex. Maybe the word I should use is mutable and immutable objects in python. But in another context, we were also suggesting these changes because of the use of a Linter, in this case, Pylint. It led us to keep almost the same structure for all the files.

diana-macrometa: Hey, I like consistency. Thanks for explaining it. :)

diana-macrometa: You have an approval, so feel free to merge this once all tests pass.

jamesm-macrometa (approved):
Looks good!

PR #199, opened 2022-08-15
Updates in code functionality

edgargarciamacrometa:
Working with documents: To keep the consistency with the API REFERENCE, I think it's a better idea if the REST endpoints also include the fabric name as f"{FED_URL}/_fabric/{FABRIC}/. In this way, the endpoints appear in the API REFERENCE in GUI. As they are right now(
  url = FED_URL + "/_api/datacenter/all") , the user who does not know the platform won't know at first sight that they can point to different fabrics.

Query Using C8QL: In the actual docs the variable url keeps pointing to the auth endpoint ( url = "{}/_open/auth".format(FED_URL)), so the variable needs to be reassigned to url = f"{FED_URL}/_fabric/_system/_api/cursor" to work properly. Also, the update call needed the WITH reserved Keyword (UPDATE c WITH). Also I added an UPSERT example

Publish Subscribe with Streams: The auth header was added for the websocket authorization (ws = create_connection(PRODUCER_URL, header=[f"authorization: {AUTH_TOKEN}"]))

Query as API: I added some artificial latency. There are moments where some race conditions occur, leading the code to fail. 

Working with graphs: In the graph traversal part, needed to be changed in the payload the vertex, from "vertex": "Jean" to "vertex": "teachers/Jean"

Stream processing: Adding the auth header for the WebSocket connection.


diana-macrometa (approved):
LGTM

jamesm-macrometa (approved):
Looks good!

PR #201, opened 2022-08-16
Update quickstart.md

milan-bluegrid:
Updated and refactored the PyC8 part, also reformatted .format() to f-strings


docs/collections/graphs/quickstart.md
```diff
@@ -227,15 +250,23 @@ An **edge collection** contains edge documents and shares its namespace with all
...
+    from c8 import C8Client
+
+    EMAIL = 'nemo@nautilus.com'
+    PASSWORD = 'xxxxx'
+    HOST = 'gdn.paas.macrometa.io'
+    GEO_FABRIC = '_system'
+
+    print("--- Connecting to C8")
+
+    client = C8Client(protocol='https', host=HOST, port=443,
```
diana-macrometa, line 262: Why do we have spaces around the equal signs elsewhere but not here?

diana-macrometa (approved):
My only comment is that we seem to be inconsistent with spacing around equal signs. Otherwise, looks good!

jamesm-macrometa (approved):
LGTM!  Agreed we should be consistent with the spacing around equal signs.

PR #219, opened 2022-08-19
Added second solution based on size of result set.

james-macrometa:
Updated the import statement to use jsc8@beta. Added second solution and revised copy to cover both solutions.


docs/collections/return-data-in-batches.md
```diff
@@ -0,0 +1,131 @@
...
+    cd jsC8
+    npm install
+    npm run dist
+```
+</TabItem>
+</Tabs>
+
+## Solutions
+
+There are two recommmended approaches depending on the volume of data being returned. The Cursor API can be used to fetch up to 30,000 records, this is an approximation based on platform limits. Attempting to return more than 30,000 from the Cursor API may cause the query to exceed the query execution time limit and raise an error. The Collection Export API is the recommended approach when retrieving an entire collection.
```
diana-macrometa: Please link to the two APIs you are referring to.

diana-macrometa: Generally we want to link to the source on the first usage. I see you link later.

Why are you linking to GitHub rather than API docs?

docs/collections/return-data-in-batches.md
```diff
@@ -0,0 +1,131 @@
...
+
+Create a connection to the GDN using the jsc8 SDK. You will need an API key, JWT, or Email/Password to complete authentication.
```
diana-macrometa:
```suggestion
Create a connection to the GDN using the jsc8 SDK. You will need an API key, JWT, or Email/Password to complete authentication. For more information, refer to [Authentication](link to authentication topic)
```

diana-macrometa: Notice that you need to add the link path here.

docs/collections/return-data-in-batches.md
```diff
@@ -0,0 +1,131 @@
...
+
+For the [Cursor API](https://github.com/Macrometacorp/jsC8/blob/master/docs/Reference/Cursor.md#cursor-api) solution, create a cursor to receive the results from your request. When calling the `query` method request options are provided as key/value pairs. It is important to set the `batchSize` value to 1000 and the `stream` value to true. This will permit the results from the query to be returned as a stream, allowing us to return subsequent batches of data.
```
diana-macrometa: These two paragraphs might make more sense as intro paragraphs under their respective h3s.

diana-macrometa (changes requested):

docs/collections/return-data-in-batches.md
james-macrometa: The code examples are using the jcs8 SDK so the syntax the user would consult to learn more about the specific methods would be in GH. I see value in including both however, what would be the correct way to add both links without creating confusion?

docs/collections/return-data-in-batches.md
james-macrometa: Added, thank you!

docs/collections/return-data-in-batches.md
james-macrometa: Moved to their respective sections.

docs/collections/return-data-in-batches.md
james-macrometa: Removed the links to the GH docs for now.

james-macrometa (commented):
@diana-macrometa please see these changes.

diana-macrometa:
If you click the circling arrows next to a person's name, then the system requests another review.

diana-macrometa (approved):
Looks good! Only thing it needs is a sidebar_position, unless you always want it to be at the very bottom of the sidebar list. Please use a two-digit number. You can check the position by running a local build.

diana-macrometa (approved):
Looks good! Feel free to merge when ready.

diana-macrometa (approved):
Once I approve it, you don't need to re-request the review.

james-macrometa:
@diana-macrometa I had to remove a file because I had changed the name and it was duplicated. Sorry for the confusion.

PR #223, opened 2022-08-24
Adding constant declarations

jesusgonzalez12:
Adding constant declarations suggested by @koshy-thomas. Thanks Koshy.

Note: This is related to PR 205.


docs/sdks/create-api-key.md
```diff
@@ -23,11 +23,14 @@ console.log('API Creation Failed: ', e);
 
 ```py
 from c8 import C8Client
+  
+API_ACTIVE = False
+CHOSEN_API_KEY_ID = "myNewKey"
+API_KEY = "my API key" # Change this to my key
+
+client = C8Client(protocol='https', host=URL, port=443, apikey = API_KEY)
```
koshy-thomas, line 34: One doubt here since this is a create apikey example shouldn't we assume that the user doesn't have an apikey yet? Hence shouldn't the C8client login method used here be based on email and password rather than apikey for "creating apikey" example? Any thoughts on this?

diana-macrometa: Yes. Assume that the user is creating an API key because they do not have one. Or we could just point them at the authentication topic and let them decide how to log in.

jesusgonzalez12: ok, thanks guys.

Suggestions have been applied.

jesusgonzalez12:
Hello team.
Please review again, approve and merge.

koshy-thomas:
> Hello team. Please review again, approve and merge.

@diana-macrometa @jamesm-macrometa can anyone of you review once more and approve if everything's fine

diana-macrometa (approved):
LGTM!

PR #225, opened 2022-08-25
Updating Graph Traversals

jesusgonzalez12:
Updating code on "Graph Traversals" section as recommended on SUP-39.


diana-macrometa (approved):
Looking good!

docs/collections/graphs/quickstart.md
```diff
@@ -453,13 +453,14 @@ A graph consists of `vertices` and `edges`. Vertices are stored as documents in
...
-    READ_CALLS = "FOR call IN CDRcalls RETURN call"
+    READ_PEOPLE = f"FOR person IN {COLLECTION_PEOPLE} RETURN person"
+    READ_CALLS = f"FOR call IN {COLLECTION_CALLS} RETURN call"
     PERSON = "Lou Feaveer"
-    GRAPH_TRAVERSAL_1 = f"FOR c IN CDRpeople FILTER c.full_name == \"{PERSON}\" FOR v IN 1..1 INBOUND c CDRcalls RETURN v"
-    GRAPH_TRAVERSAL_2 = f"FOR c IN CDRpeople FILTER c.full_name == \"{PERSON}\" FOR v IN 1..1 OUTBOUND c CDRcalls RETURN v"
+    GRAPH_TRAVERSAL_1 = (f"FOR c IN {COLLECTION_PEOPLE} FILTER c.full_name == \"{PERSON}\"
+        f"FOR v IN 1..1 INBOUND c {COLLECTION_CALLS} RETURN v")
+    GRAPH_TRAVERSAL_2 = (f"FOR c IN {COLLECTION_PEOPLE} FILTER c.full_name == \"{PERSON}\"
+        f"FOR v IN 1..1 OUTBOUND c {COLLECTION_CALLS} RETURN v")
```
koshy-thomas, line 463: Two f-strings are used in GRAPH_TRAVERSAL_1 AND GRAPH_TRAVERSAL_2 by mistake it seems you can remove the second f-string after \"{PERSON}\" in line 461 and 463

koshy-thomas:
Also, in Line 450  we need to change the global URL from "gdn.pass" to "gdn.paas", a typo

docs/collections/graphs/quickstart.md
jesusgonzalez12: Hi @koshy-thomas. Using two lines of f-strings is the right thing to do. Keeping it in a single line makes the line over 100 characters. This is to comply with PEP8.

jesusgonzalez12: The typo on line 450, "pass" to "paas" has been fixed.
Thanks @koshy-thomas.

koshy-thomas: The problem there is that while you have used two f-strings you haven't closed the first f-string in both GRAPH_TRAVERSAL_1 and GRAPH_TRAVERSAL_2

koshy-thomas: So just adding a double quote after \"{PERSON}\" would close the first f-string

jesusgonzalez12: You're right :)

I probably got confused with the "\ pairs. I've seen those more commonly with API calls, not so much with SDK's.
I just removed them.
Thanks @koshy-thomas

koshy-thomas: Nope we need those \\" pairs for c8ql string comparison we just want to add another double quote after \\"{PERSON}\\".... so it will be \\"{PERSON}\\""

jesusgonzalez12: Thanks again @koshy-thomas.

I've made the changes. I think I've seen the \" pair not being applied in the SDK and working. That's an exercise I'll try to replicate for personal reference.

jesusgonzalez12: Thanks for the peer-review and feedback @koshy-thomas. We saw that the problem presents when there is a value expected to have quotes or double quotes in C8QL, in this instance, being compared.
Essentially, if using the UI and the query expects quotes or double quotes, they must be inserted, regardless if we're using a simple string or a variable/constant with a type of string.

Thanks.
Code has been updated.
We can move on now to the next step.

PR #226, opened 2022-08-25
Changes for updateItem DynamoDb

koshy-thomas:
Changing UPDATED_NEW and UPDATED_OLD to RETURN_NEW (so that user never gets confused whether the remaining attributes have been overwritten or don't exist because we had a previous such case where we were overwriting the not updated values)

Also corrected the update expression in usingaws-jsnodejs example.


diana-macrometa (approved):
Nice change!

PR #229, opened 2022-09-01
changed JS connection argument to apiKey.

james-macrometa:
The token will only work with a JWT but the first step instructs the creation of an API Key. This method should work as expected.


diana-macrometa (approved):
Nice catch!

PR #233, opened 2022-09-07
Update create-document-store.md

diana-macrometa:
Added tabs with CLI and API links to Create a Document Store.


jamesm-macrometa:
Looks good!

PR #235, opened 2022-09-08
Create document-store-cli.md

diana-macrometa:
While working on other stuff, I realized that we are missing an entire family of CLI commands. This PR imports the `gdnsl document` commands from the original CLI doc repo.


jamesm-macrometa:
Thanks @diana-macrometa !  Good catch.

jamesm-macrometa (approved):

PR #257, opened 2022-09-19
Updated code and made it standard compliant

milan-bluegrid:
Made the code standard compliant and added constants before function calls, also moved the login details in the async function because await cannot work outside async.


diana-macrometa (approved):
LGTM

PR #265, opened 2022-09-21
Removing CLI command information

Luba89:
CLI command is showing a version of the GDNSL; currently, there is no command that will show the version of the platform.


diana-macrometa (approved):
Please update this in the 17.10 release notes as well.

PR #266, opened 2022-09-21
Updated code and made it standard compliant

milan-bluegrid:
Removed the username and tenant in the authentication part because there is no need for them to be entered as part of the authentication.


diana-macrometa (approved):
LGTM

PR #267, opened 2022-09-21
Doc 458 explain cursors

ghost:
- Added a more detailed explanation of what the cursor is
- Corrected code samples (Fist example was not working, the second example change is just syntax sugar to be aligned with the previous)


diana-macrometa (approved):
This is a good start. I think we also need a topic in the Queries and Query Workers section explaining what a cursor is. That topic could then link to this and to CLI commands, API commands, and so on.

I suggest a separate topic for that one.

docs/collections/returning-collection-data.md
```diff
@@ -43,28 +43,23 @@ From source,
...
+We can create a cursor to receive the results from our request. A cursor details where we left off when reading a stream of responses from API, so we can continue to read the batch of required data. When calling the `query` method request options are provided as key/value pairs. For a detailed explanation of options please visit [Api Reference](https://macrometa.com/docs/api#/operations/createQueryCursor). It is important to set the `batchSize` to required value (code ex. 1000) and the `stream` value to true. This will permit the results from the query to be returned as a stream, allowing us to return subsequent batches of data. In response, the entire result set is loaded into the cursor. (code ex. `cursor`) Therefore we must be mindful of client-side memory capacity when running queries that can potentially return a large result set.
```
jamesm-macrometa:
```suggestion
We can create a cursor to receive the results from our request. A cursor details where we left off when reading a stream of responses from API, so we can continue to read the batch of required data. When calling the `query` method request options are provided as key/value pairs. For a detailed explanation of options visit the [API reference](https://macrometa.com/docs/api#/operations/createQueryCursor). Set the `batchSize` to the required value (code ex. 1000) and the `stream` value to true. This will permit the results from the query to be returned as a stream, allowing us to return subsequent batches of data. In response, the entire result set is loaded into the cursor (code ex. `cursor`). Therefore we must be mindful of client-side memory capacity when running queries that can potentially return a large result set.
```

A few minor grammatical/concision suggestions here.  Also, no need to say "please" in technical docs.

Otherwise looks good!

jamesm-macrometa (approved):

docs/collections/returning-collection-data.md
ghost: Thank you @jamesm-macrometa! 👍

ghost:
Thanks, @diana-macrometa, and @jamesm-macrometa. Let's continue 🚀

ghost:
What is the procedure after review? Can I merge to the main? 😅 @diana-macrometa

diana-macrometa (approved):
Good to go once you address James' suggestions.

PR #285, opened 2022-09-29
Remove or hide this page

Luba89:
Currently, users can't see or change any limits on our platform; I think this page should be removed.


diana-macrometa (changes requested):
@mikeridley, @elof, @durgagokina, need some context here. 

Are we planning on doing anything with these limits in the future? Is this information useful to users at all?

Options are:
1. Leave this topic as-is.
2. Delete this topic.
3. Hide this topic but leave it here so we can unhide it later.

diana-macrometa:
@Luba89, please answer the questions.

Luba89:
> @Luba89, please answer the questions.

@diana-macrometa Users can't change the Stream Workers limits and can't see the page for changing limits. I would remove this page, but we can save it as an internal doc.

diana-macrometa (approved):
Sound good to me. Please make sure this content is saved somewhere that people who need it can access it.

PR #289, opened 2022-09-30
This should be internal

Luba89:
This is not something that regular users can change, maybe this will change with the new provision approach, but currently, the GDN users can't see or change this setting. This can only access by the mm admin account.


diana-macrometa (approved):
Okeydokey, though if it is useful for internal folx, then I recommend putting the content in Notion or the internal KB when it is up.

PR #299, opened 2022-10-05
New Favicon and Icon for README

dustinlarimer:


dustinlarimer:
The icon added to the `README` will load from the deployed docs site, meaning it will be momentarily broken until the site deploys after merge.

PR #311, opened 2022-10-11
Streamlined account management

jamesm-macrometa:
Streamlined account management by moving the first two steps into the intro.

### DIFF
--- a/docs/account-management/users/index.md
+++ b/docs/account-management/users/index.md
@@ -5,12 +5,15 @@ title: Create and Manage Users
 
 This section describes how to create and manage user accounts.
 
+Follow these instructions to access the account management screen:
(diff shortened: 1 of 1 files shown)


diana-macrometa:
I do not like this change, it's not intuitive to look for steps in the intro rather than the procedure.

A better solution might be to use snippets for the steps that we keep repeating, such as logging in to accounts and such.

jamesm-macrometa:
> I do not like this change, it's not intuitive to look for steps in the intro rather than the procedure.
> 
> A better solution might be to use snippets for the steps that we keep repeating, such as logging in to accounts and such.

Sounds good, I'll rework this today and re-open the PR.

PR #318, opened 2022-10-13
Add API key authentication instructions

diana-macrometa:
- Added a task explaining how to authenticate with SDKs and an API key.
- Updated the auth example with same.

**Reviewers** - If you could provide instructions for authenticating with an API key via API, that would be great.


docs/account-management/auth/connect-to-gdn.md
```diff
@@ -60,11 +62,16 @@ main()
...
 from c8 import C8Client
+
+# Authenticate with your email and password
 client = C8Client(protocol='https', host='gdn.paas.macrometa.io', port=443,
                     email='nemo@nautilus.com', password='xxxxx',
                     geofabric='_system')
 
-# Or using token
+# Authenticate with API key (recommended for apps)
+client = C8Client(protocol='https', host=URL, port=443, apikey = API_KEY, geofabric = GEO_FABRIC)
```
koshy-thomas, line 77: the variables URL, API_KEY, and GEO_FABRIC are not defined here.

docs/account-management/auth/api-key-auth.md
```diff
@@ -0,0 +1,48 @@
...
+</TabItem>
+<TabItem value="py" label="Python">
+
+```py
+from operator import concat
+import base64
+import json
+import warnings
+from c8 import C8Client
+import six
```
koshy-thomas: We don't need all these imports here (since it's just creating a client instance)
 For that `from c8 import C8Client `should be enough

docs/account-management/auth/api-key-auth.md
```diff
@@ -0,0 +1,48 @@
...
+warnings.filterwarnings("ignore")
```
koshy-thomas: warnings.filterwarnings("ignore") is also not needed here

koshy-thomas (commented):
This is a sample python code demonstrating how to use apikey to make REST API calls:

import requests

\# Constants

FEDERATION = "api-gdn.paas.macrometa.io"
FED_URL = f"https://{FEDERATION}"
FABRIC = "_system"
API_KEY = "XXXXX" # To be replaced by your apikey

\# apikey keyword needs to be appended to the apikey
AUTH_TOKEN = f"apikey {API_KEY}"

\# Create a HTTPS Session authenticated with apikey
session = requests.session()
session.headers.update({"content-type": 'application/json'})
session.headers.update({"authorization": AUTH_TOKEN})

\# Now you can use the above session to make any API request
\# For example we are calling the 'List all collections' API below
url = f"{FED_URL}/_fabric/{FABRIC}/_api/collection"

response = session.get(url)
print(response.text)`

koshy-thomas (commented):
This a sample Javascript code to create REST API requests via apikey :

class APIRequest {
  _headers = {
    Accept: "application/json",
    "Content-Type": "application/json"
  };

  constructor (url, apiKey) {
    this._url = url;
    this._headers.authorization = \`apikey ${apiKey}\`; // apikey keyword needs to be appended
  }

  _handleResponse (response, resolve, reject) {
    if (response.ok) {
      resolve(response.json());
    } else {
      reject(response);
    }
  }

  req (endpoint, { body, ...options } = {}) {
    const self = this;
    return new Promise(function (resolve, reject) {
      fetch(self._url + endpoint, {
        headers: self._headers,
        body: body ? JSON.stringify(body) : undefined,
        ...options
      }).then((response) => self._handleResponse(response, resolve, reject));
    });
  }
}

const federationName = "api-gdn.paas.macrometa.io";
const federationUrl = \`https://${federationName}\`;
const fabric = "_system";
const apiKey = "XXXXX"; // your apikey

const run = async function () {
  try {
    const connection = new APIRequest(federationUrl, apiKey);
    // Now we can use the above connection to call APIs
    // For example we are calling the 'List all collections' API below
    const response = await connection.req(\`/_fabric/${fabric}/_api/collection\`, {
      method: "GET"
    });
    console.log(response)
  } catch (e) {
    console.error(e);
  }
};

run();

docs/account-management/auth/connect-to-gdn.md
```diff
@@ -60,11 +62,21 @@ main()
...
+
+# Authenticate with JWT
 client = C8Client(protocol='https', host='gdn.paas.macrometa.io', port=443, token="XXXX")
```
koshy-thomas: since now that we have defined the variable URL here so instead of gdn.paas.macrometa.io we should use that variable and maybe you can also add the fabric too, i.e. you can add `geofabric=GEO_FABRIC` here too just like in the line just above this

docs/account-management/auth/connect-to-gdn.md
```diff
@@ -60,11 +62,21 @@ main()
...
+
+# Define constants
+URL = "gdn.paas.macrometa.io"
+GEO_FABRIC = "_system" # Change this if you need to log in to a fabric other than _system.
+API_KEY = "my API key" # Change this to your API key.
+
+# Authenticate with your email and password
 client = C8Client(protocol='https', host='gdn.paas.macrometa.io', port=443,
                     email='nemo@nautilus.com', password='xxxxx',
                     geofabric='_system')
```
koshy-thomas: since now that we have defined the variables URL, and GEO_FABRIC here so instead of gdn.paas.macrometa.io  and "_system" we can use the same variable

diana-macrometa:
How does this look? Ready for approval?

docs/account-management/auth/api-key-auth.md
```diff
@@ -0,0 +1,129 @@
...
+session = requests.session()
+session.headers.update({"content-type": 'application/json'})
+session.headers.update({"authorization": AUTH_TOKEN})
+
+# Now you can use the above session to make any API request
+# For example we are calling the 'List all collections' API below
+url = f"{FED_URL}/_fabric/{FABRIC}/_api/collection"
+
+response = session.get(url)
+print(response.text)`
```
koshy-thomas:
```suggestion
print(response.text)
```

docs/account-management/auth/api-key-auth.md
```diff
@@ -0,0 +1,129 @@
...
+```
+
+Sample Javascript code demonstrating how to use API key to make REST API calls:
+
+```js
+class APIRequest {
+_headers = {
+Accept: "application/json",
+"Content-Type": "application/json"
+};
```
koshy-thomas: Indentation seems to be lost in this javascript code, most likely because of github comments.. Sent you the indented version of the code on slack

docs/account-management/auth/connect-to-gdn.md
```diff
@@ -95,11 +107,11 @@ headers = {
 response = requests.post(url, data = json.dumps(payload), headers = headers)
 
 if response.status_code == 200:
```
diana-macrometa, line 109: @koshy-thomas, is the indentation right here?

koshy-thomas: Indeed it was wrong, you have corrected the below part already, I just added a suggestion on the above part

docs/account-management/auth/connect-to-gdn.md
```diff
@@ -82,7 +94,7 @@ EMAIL = "nemo@nautilus.com"
 PASSWORD = "xxxxxx"
 AUTH_TOKEN = "bearer "
 
-# Create a HTTPS Session
+# Create an HTTPS Session
 url = "{}/_open/auth".format(FED_URL)
 payload = {
 'email':EMAIL,
```
koshy-thomas, line 100:
```suggestion
  'email':EMAIL,
```

diana-macrometa: DAmn, I knew I should have waited. I'll put these changes in a new PR.

koshy-thomas: same needs to be applied for password (the below line) and 'content-type' (line no. 104)

PR #362, opened 2022-11-01
Subscribe and Unsubscribe from Streams

diana-macrometa:
Created the following topics:

- Subscribe to Streams
- Unsubscribe from Streams

**Reviewers**
- I could not find a way to subscribe to streams via the CLI, so there is no tab for that.
- I'm not confident that I grabbed the right code blocks for the Subscribe to Streams API-JS tab, please check that carefully.


docs/streams/subscribe-streams.md
```diff
@@ -0,0 +1,270 @@
+---
+sidebar_position: 70
+title: Subscribe to Streams
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+This page explains how to publish messages to a stream in Macrometa.
```
koshy-thomas: We are not "publishing" but subscribing here right? ..
Also yes I didn't find any subscribe to stream CLI command too..

docs/streams/subscribe-streams.md
```diff
@@ -0,0 +1,270 @@
...
+
+<Tabs groupId="operating-systems">
+<TabItem value="py" label="Python SDK">
+
+You must [Install the Python SDK](../sdks/install-sdks.md) before you can run this code.
+
+```py
+from operator import concat
```
koshy-thomas: concat is not needed but json and base64 is needed
```suggestion
import json
import base64
```

docs/streams/subscribe-streams.md
```diff
@@ -0,0 +1,270 @@
...
+from c8 import C8Client
+
+# Connect to GDN.
+URL = "gdn.paas.macrometa.io"
+GEO_FABRIC = "_system"
+API_KEY = "xxxxxx" # Change this to your API key
+is_local = False # For a global stream pass True and False for local stream
```
koshy-thomas:
```suggestion
is_local = False # For a global stream pass False and True for a local stream
```

docs/streams/subscribe-streams.md
diana-macrometa: Thanks for catching that!

docs/streams/subscribe-streams.md
```diff
@@ -0,0 +1,268 @@
...
+    // Send message acknowledgement
+    consumer.send(JSON.stringify({ messageId }));
+  });
+})();
+```
+
+</TabItem>
+<TabItem value="api-py" label="REST API - Python">
+
+Use our interactive API reference with code generation in 18 programming languages to [publish a message](https://macrometa.com/docs/api#/operations/PublishStreamMessage).
```
koshy-thomas: @diana-macrometa Is the publish a message reference intentional here meaning.. telling the user to use that PublishSTreamMessage api and then test our subscribe code...
Or I am overthinking and this was just a mistake since the page is subscribe stream? (Also this thing is repeated in the file so check all the instances)

diana-macrometa: No, it's a copy/paste error, I must have been tired. Could have sworn I updated all those, but obviously not. ARGH. I will fix this.

docs/streams/unsubscribe-streams.md
```diff
@@ -0,0 +1,118 @@
...
+# Delete subscription
+url = f"{HTTP_URL}/_fabric/{FABRIC}/_api/streams/{stream_type}s.{STREAM_NAME}/subscriptions/{CONSUMER_NAME}?global=true"
+resp = session.delete(url)
+print("Subscription deleted: ", resp.text)
+```
+
+</TabItem>
+<TabItem value="api-js" label="API - JS">
+
+Use our interactive API Reference with code generation in 18 programming languages to [Delete All Subscriptions](https://macrometa.com/docs/api#/operations/DeleteAllSubscriptions) from a stream or [Remove a Subscription](https://macrometa.com/docs/api#/operations/DeleteSubscription).
```
koshy-thomas, line 152: @diana-macrometa I think there is a major misunderstanding here... It is not deleting all subscription from a stream... Unsubscribe API operation means removing the given subscription from all streams either global or local) which is defined by a "global" parameter which we give as input to the API.. I think it would be best if you login to smoke1 or smoke3 and check the API references from the UI that should clear the picture for you..

And then the second one delete subscription is as it is like just delete a given subscription from a given stream
But the deleteallsubscriptions thing is wrong.. plus our apidocs on docs repo seems not updated from a while.. (like I don't see the "global" parameter in unsubscribe method of api ref in docs but it is present any GDN newer versions


and this is for all tab items of this file

docs/streams/unsubscribe-streams.md
```diff
@@ -0,0 +1,118 @@
...
+
+This page explains how you can unsubscribe from streams in Macrometa.
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+<Tabs groupId="operating-systems">
+<TabItem value="api-py" label="API - Python">
+
+Use our interactive API Reference with code generation in 18 programming languages to [Delete All Subscriptions](https://macrometa.com/docs/api#/operations/DeleteAllSubscriptions) from a stream or [Remove a Subscription](https://macrometa.com/docs/api#/operations/DeleteSubscription).
```
koshy-thomas, line 101: The same comment regarding deleteallsubscriptions confusion.. this is on all tab items

docs/streams/unsubscribe-streams.md
diana-macrometa: Theoretically, the API docs are built from the same Swagger .json file. I don't think they update except when a release goes out.

koshy-thomas: Even In GDN you can see the global parameter under Unsubscribe Stream API if you go to API reference in UI

koshy-thomas: I see that this file:
https://github.com/Macrometacorp/docs/tree/master/static/swagger
hasn't been updated in more than2 months

diana-macrometa: LOL. Bold of you to assume I know what it meant!

koshy-thomas (approved):

PR #363, opened 2022-11-02
SDK fix

Luba89:
Python was missing a few libs and had one that the script didn't use. Js part had part of code that is not necessary for this example.

### DIFF
--- a/docs/streams/subscribe-streams.md
+++ b/docs/streams/subscribe-streams.md
@@ -14,8 +14,9 @@ This page explains how to publish messages to a stream in Macrometa.
 You must [Install the Python SDK](../sdks/install-sdks.md) before you can run this code.
 
 ```py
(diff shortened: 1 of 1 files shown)


diana-macrometa (approved):

diana-macrometa:
I think this conflicts with some of the changes @koshy-thomas made. Please reconcile.

koshy-thomas:
@Luba89 I guess your major concerns of missing lib and additional function are already being addressed via https://github.com/Macrometacorp/docs/pull/362/files.. Do let me know if it is okay to close your PR since I have some more things for to fix/add on 362 PR...

Luba89:
@koshy-thomas yup, you can close it. Also we can delete :console.log("Authentication done!!...");
as we don't do any auth in line before. It is just an object creation command.

PR #379, opened 2022-11-07
Delete note

Luba89:
Search becomes GA in all tiers, including the PLAY tier


diana-macrometa (approved):
Nice catch, Luba! Thanks!

PR #384, opened 2022-11-08
Update create-streams.md

bdzekic1:
Changed comment sign on line 78 from `#` to `//` since `#` is for PY but the comment is inside JS section


diana-macrometa (approved):
Good catch, thank you!

PR #386, opened 2022-11-08
Update getting-started-with-streams.md

bdzekic1:
Removed part of the code which can create potential confusion to the users since it's giving a sense that connection on the receiver end can be closed through senders' input, but that behavior is not present in the current code, so until the proper implementation of that behavior, this code block should be missing.


diana-macrometa:
Adding @koshy-thomas, because he wrote that part of the code.

koshy-thomas (approved):

PR #388, opened 2022-11-10
Doc 446 - Rework Stream Worker SDK example

diana-macrometa:
This PR improves the Stream Worker SDK Example and makes minor adjustments to Getting Started with Streams.

**Reviewers**

I need some help with this. There are a bunch of problems.

- I had problems getting the ad hoc query to work with the Python example when it was part of the larger process, so I removed it from the full demo.
- When running the full demo example, the ad hoc query doesn't work. It seems to run after the stream is deleted, so fails.
- The user should be able to add the parts of the example one at a time to a .js or .py file and run them after each step successfully, preferably with viewable results. This is not working at all with the JavaScript, it waits for a Promise that never comes.
- The JS full demo totally does not match the way the individual steps are written, which will confuse the user. The full demo at the end should be very similar or identical to what you get when you compile the steps, and differences need to be explained.
- Similar problem with the Python version, but it's not as bad.


docs/cep/stream-workers-sdk-example.md
```diff
@@ -308,164 +313,176 @@ print("Activate", client.activate_stream_app('Sample-Cargo-App', True))
...
-    INSERT INTO SampleCargoAppDestStream
-    SELECT weight
-    FROM SampleCargoAppInputTable;
-    """
-# Validate a stream application
+-- Define Stream SampleCargoAppDestStream
+CREATE SINK SampleCargoAppDestStream WITH (type = 'stream', stream = "SampleCargoAppDestStream", replication.type="local") (weight int);
+
+-- Defining a Destination table to dump the data from the stream
+CREATE STORE SampleCargoAppDestTable WITH (type = 'database', stream = "SampleCargoAppDestTable") (weight int);
```
StoyanTc: Let's make this definition more simple:
```
CREATE TABLE SampleCargoAppDestTable (weight int);
````

**Review by StoyanTc — CHANGES_REQUESTED** [review]
I have tested the full demo with pyC8. It's working with ad-hock query, the only requirement in such case is to have the stream worker active (published). Additionally I think would be better to change the table's definition from 
```CREATE STORE SampleCargoAppDestTable WITH (type = 'database', stream = "SampleCargoAppDestTable") (weight int);``` 
to 
```CREATE TABLE SampleCargoAppDestTable (weight int);```

File: docs/cep/stream-workers-sdk-example.md (hunk 1 above)
Comment (koshy-thomas) [reply docs/cep/stream-workers-sdk-example.md]: @StoyanTc can you check my comment on https://macrometa.atlassian.net/browse/CHG-605. This is relevant to this example and all examples where we are not specifying `replication.type` in `stream workers` Since CREATE STORE, TABLE, SINK, SOURCE, without specifying the replication type will create a `local `collection/stream and we have removed support for local collections from play and paas and might remove support from other federation as well..

Comment (diana-macrometa) [reply docs/cep/stream-workers-sdk-example.md]: @StoyanTc, would you mind either adding the change as a suggestion or putting it in as a commit? I'm not super confident in my command of stream workers or JavaScript.

Btw, if you want to change this example, it's used several times in the docs and I think it is also in the sample apps in the product.

Comment (diana-macrometa) [reply docs/cep/stream-workers-sdk-example.md]: @StoyanTc, you say you tested the full demo, but did you also test the per-step method I explained in the comment?

Comment (StoyanTc) [reply docs/cep/stream-workers-sdk-example.md]: @diana-macrometa - I don't have rights to make changes here. 
There are two reasons for the issue you have with the ad-hock query in step 5:
1. The original sample `Sample-Cargo-App` does not have store/table `SampleCargoAppDestTable`.
2. The created stream worker is not activated/published.

**Comment by koshy-thomas** [comment]
@diana-macrometa can you check now if everything is as per the expected lines and is working fine?
I have updated the example codes as per the expectations mentioned by you but please check thoroughly once and let me know if there is any problem.

**Review by koshy-thomas — APPROVED** [review]

**Comment by diana-macrometa** [comment]
@koshy-thomas, I got stopped at step 3. It doesn't like the `await`. Removing the `await` gave it a different error, `appDefinition is not defined`.

```
(base) dianapayton@Dianas-MacBook-Pro js % node stream-worker-sdk.js
/Users/dianapayton/Documents/js/stream-worker-sdk.js:70
      result = await client.createStreamApp(dclist, appDefinition);
               ^^^^^

SyntaxError: await is only valid in async functions and the top level bodies of modules
    at Object.compileFunction (node:vm:360:18)
    at wrapSafe (node:internal/modules/cjs/loader:1088:15)
    at Module._compile (node:internal/modules/cjs/loader:1123:27)
    at Module._extensions..js (node:internal/modules/cjs/loader:1213:10)
    at Module.load (node:internal/modules/cjs/loader:1037:32)
    at Module._load (node:internal/modules/cjs/loader:878:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:81:12)
    at node:internal/main/run_main_module:23:47
    ```

diana-macrometa:
Here's the code minus login:
```
// This is a supporting function
function messageHandler (error) {
    const message = {
      "StatusCode ": error.statusCode,
      "ErrorMessage ": error.message,
      "ErrorNum ": error.errorNum
    };
    console.log(message);
  }
  
  // This is the main function
  async function main () {
    try {
      // Define the stream app to validate.
      const appDefinition = `
        @App:name('Sample-Cargo-App')
        @App:qlVersion("2")
        @App:description('Basic stream worker to demonstrate reading data from input stream and store it in the collection. The stream and collections are automatically created if they do not already exist.')
        /**
        Test the stream worker:
            1. Open Stream SampleCargoAppDestStream in console. The output can be monitored here.
            2. Upload following data into SampleCargoAppInputTable collection:
                {"weight": 1}
                {"weight": 2}
                {"weight": 3}
                {"weight": 4}
                {"weight": 5}
            3. Following messages are shown on the SampleCargoAppDestStream Stream Console:
                [2021-08-27T14:12:15.795Z] {"weight":1}
                [2021-08-27T14:12:15.799Z] {"weight":2}
                [2021-08-27T14:12:15.805Z] {"weight":3}
                [2021-08-27T14:12:15.809Z] {"weight":4}
                [2021-08-27T14:12:15.814Z] {"weight":5}
        */
  
        -- Create Table SampleCargoAppInputTable to process events.
        CREATE SOURCE SampleCargoAppInputTable WITH (type = 'database', collection ="SampleCargoAppInputTable", collection.type="doc", replication.type="global", maptype='json') (weight int);
  
        -- Create Stream SampleCargoAppDestStream
        CREATE SINK SampleCargoAppDestStream WITH (type = 'stream', stream ="SampleCargoAppDestStream", replication.type="local") (weight int);
  
        -- Data Processing
        @info(name='Query')
        INSERT INTO SampleCargoAppDestStream
        SELECT weight
        FROM SampleCargoAppInputTable;`
  
      // Validate the stream worker code.
      console.log("--- Validating stream worker definition");
      let result;
      result = await client.validateStreamApp(appDefinition);
      console.log(!result.error)
    } catch (e) {
      console.log(messageHandler(e));
    }
  }
  main();

      // The stream app will be created by default in the local region. Optionally, you can use dclist to deploy stream
    // app in all / selected regions
    const dclist = []
    console.log("--- Creating stream worker");
    try {
      result = await client.createStreamApp(dclist, appDefinition);
      console.log(!result.error);
    } catch (e) {
      if (e.statusCode === 409) {
        console.log("Stream worker already exists");
      } else {
        throw e;
      }
    }
```

koshy-thomas:
@diana-macrometa you need to add the createStream app code in side the try block of main function

koshy-thomas:
I have mentioned the same as instruction too

diana-macrometa:
I got this when I tested the full JS demo:
```
(base) dianapayton@Dianas-MacBook-Pro js % node stream-worker-sdk.js
--- Validating stream worker definition
true
--- Creating stream worker
Stream worker already exists
--- Activating `Sample-Cargo-App`
true
--- Updating stream worker `Sample-Cargo-App`
true
--- Waiting 10 seconds for all the resources to be ready
--- Inserting data to `SampleCargoAppInputTable` collection
{
  'StatusCode ': 409,
  'ErrorMessage ': "Each document in a collection must have a unique key, and the provided one is already used in collection '_userqueries'.",
  'ErrorNum ': 1210
}
undefined
``

diana-macrometa:
> @diana-macrometa you need to add the createStream app code in side the try block of main function

Ah ha! I'll update the instructions and try again.

koshy-thomas:
image

koshy-thomas:
I will connect with you on slack best

PR #405, opened 2022-11-21
Update consuming-data.md

bdzekic1:
one `tip` and `info` section had `!!!` instead of `:::`, which didn't do anything, and just looked uncompleted


bdzekic1:
I haven't changed the code completely. I'm able to see it now that I haven't placed ending columns `:::`. Also, under the section `Consume messages in the custom format` at the end of the section, it's not reading `:::tip` properly, and the text is undefined

bdzekic1:
Some issues occur when editing sections I want to commit, so I will have to inspect further to be able to make proper PR cause I'm not that familiar with the `:::` type and getting weird behavior, hence deleting this branch

PR #411, opened 2022-11-28
DOC-474: Update Stream producers

diana-macrometa:
This PR is a review and edit of Producers in the Stream section.

@StoyanTc and @dlozina-macrometa , I have a lot of questions based on the existing content.

- We describe two send modes, but we don't talk about where we set them or if we need to set them on the Macrometa side. How does the send mode affect the streams? How do we set them?
- We say we support two types of compression, but where do we set that in the streams? What do users need to do?
- Batching. How do we enable batching and set the max. # of messages and max. publish latency to define the batch size?


cepstreams-macrometa:
> This PR is a review and edit of Producers in the Stream section.
> 
> @StoyanTc and @dlozina-macrometa , I have a lot of questions based on the existing content.
> 
> * We describe two send modes, but we don't talk about where we set them or if we need to set them on the Macrometa side. How does the send mode affect the streams? How do we set them?
> * We say we support two types of compression, but where do we set that in the streams? What do users need to do?
> * Batching. How do we enable batching and set the max. # of messages and max. publish latency to define the batch size?

@diana-macrometa - I think we need to provide samples. However I'm not sure whether `async` mode is supported by our drivers - `pyC8` and `jsC8`. It's the same for the compression. The `batching` is on by default but again we should check how to setup batch size when use our drivers. 
Alternative way is just to provide samples with Pulsar clients.

diana-macrometa:
If you want to provide samples, that's cool. I know Pulsar exists, but that's the limit of my Pulsar expertise. 

If there are inaccuracies in the article, like saying we support things that we do not, please add suggestions or a commit to fix it. That is the purpose of this review, after all.

StoyanTc:
> If you want to provide samples, that's cool. I know Pulsar exists, but that's the limit of my Pulsar expertise.
> 
> If there are inaccuracies in the article, like saying we support things that we do not, please add suggestions or a commit to fix it. That is the purpose of this review, after all.

There are not inaccuracies according to me. Probably we could commit the current doc and add the samples later? What do you think?

diana-macrometa:
It's a good policy not to tell someone that they need to do a thing or can do a thing (like have different modes or set up readers) without telling them how to do the thing. If I could find anything in the API or CLI that worked, I would just link to them. The SDK had ways to set the subscriptions, maybe they have ways to set this stuff, but I don't know how to dig that out of the code.

The jsC8 has documentation on [cursors](https://github.com/Macrometacorp/jsC8/blob/master/docs/Reference/Cursor.md), [createStreamProducer](https://github.com/Macrometacorp/jsC8/blob/master/docs/Reference/Stream.md#clientcreatestreamproducer), and [createStreamReader](https://github.com/Macrometacorp/jsC8/blob/master/docs/Reference/Stream.md#clientcreatestreamreader) that seem to have the right parameters. Looks like most of the same might be in pyC8 [StreamCollection](https://pyc8.readthedocs.io/en/latest/specs.html#streamcollection) as well. Can you or @dlozina-macrometa write code blocks for the things that need instructions? Doesn't need to be a whole sample app, just a few lines that the user can copy/paste.

diana-macrometa:
I'll make some Jira tickets to capture next steps.

StoyanTc:
@diana-macrometa - I have added examples as comments within the ticket.

diana-macrometa:
Thanks, I will take a look.

PR #412, opened 2022-11-28
DOC-474 Update Stream consumers.md

diana-macrometa:
This PR is a review and edit or Consumers in the Stream section.

@StoyanTc and @dlozina-macrometa, I have some questions based on the existing content.

- It says there are various receive modes, but how and where do we define them in the stream? I'm guessing there is a setting?
- How do I set the acknowledgement type? We mention two different acknowledgement types, but not how or where to set them.
- How do I set up a reader? And how do I tell it where to begin processing messages?
- How doe I set a cursor on the stream?


StoyanTc:
> This PR is a review and edit or Consumers in the Stream section.
> 
> @StoyanTc and @dlozina-macrometa, I have some questions based on the existing content.
> 
> * It says there are various receive modes, but how and where do we define them in the stream? I'm guessing there is a setting?
> * How do I set the acknowledgement type? We mention two different acknowledgement types, but not how or where to set them.
> * How do I set up a reader? And how do I tell it where to begin processing messages?
> * How doe I set a cursor on the stream?

@diana-macrometa - This could be done by samples with Pulsar clients. I'm not sure `pyC8` and `jsC8` support those options.

diana-macrometa:
@StoyanTc, it says we do in the topic. Can you add suggestions or a commit to remove what we do not support?

PR #414, opened 2022-12-01
Fixing some typos & asking questions

jesusgonzalez12:
Aside from fixing a few typos, I changed "StreamApp" reference to "Stream Worker". If you consider it should be "stream worker" is better, go ahead.

A few other questions I have is:
1) I did not understand the line "Here, the keys of payload mapping can be defined using the dot notation as a.b.c, or using any constant string value as '$abc'.", and I cannot see an example of that use case. If this goes here, an example would greatly help. 2) I think the line  "unavailable or coursing errors" should be "unavailable or causing errors".
3) We have this in one query: "CREATE SOURCE <source_name> WITH (type = 'source_type'". Shouldn't it be "... WITH (source.type = ..."?


diana-macrometa:
@StoyanTc, can you answer the questions in the description?

docs/cep/query-guide/stream.md
```diff
@@ -68,8 +68,8 @@ The following is the list of source types supported by Stream:
 
 |Source type | Description|
 | ------------- |-------------|
-| `database` | Allow stream app to consume events from collections (doc, graphs) running in the same or different geo fabric. |
-| `stream` | Allow stream app to consume events from streams (local, geo-replicated) running in the same or different geo fabric. |
+| `database` | Allow the Stream Worker to consume events from collections (doc, graphs) running in the same or different geo fabric. |
```
diana-macrometa: Do not capitalize stream worker unless you are referring to the UI, such as "click **Stream Worker**.

docs/cep/query-guide/stream.md
```diff
@@ -227,7 +227,7 @@ Distributed sink provides a way to publish Stream events to multiple endpoints i
 
 **Syntax**
```
diana-macrometa: Change these bold lines pretending to be headings into actual headings.

docs/cep/query-guide/stream.md
```diff
@@ -419,7 +419,7 @@ Stream infers and automatically defines the fault stream of `TempStream` as give
 CREATE STREAM !TempStream (deviceID long, roomNo int, temp double, _error object);
 ```
 
-The StreamApp extending the above the use-case by adding failure generation and error handling with the use of [queries](#query) is as follows.
+The Stream Worker extends the above use-case by adding failure generation and error handling with the use of [queries](#query) is as follows.
 
 Note: Details on writing processing logics via [queries](#query) will be explained in later sections.
```
diana-macrometa:
```suggestion
:::note
Details on writing processing logics via [queries](#query) will be explained in later sections.
:::
```

docs/cep/query-guide/stream.md
```diff
@@ -419,7 +419,7 @@ Stream infers and automatically defines the fault stream of `TempStream` as give
 CREATE STREAM !TempStream (deviceID long, roomNo int, temp double, _error object);
 ```
 
-The StreamApp extending the above the use-case by adding failure generation and error handling with the use of [queries](#query) is as follows.
+The Stream Worker extends the above use-case by adding failure generation and error handling with the use of [queries](#query) is as follows.
```
diana-macrometa:
```suggestion
The stream worker extends the above use case by adding failure generation and error handling with the use of [queries](#query) as follows.
```

diana-macrometa (changes requested):

StoyanTc:
> @diana-macrometa @jesusgonzalez12  - 
1. I cannot understand your question, please provide a link to which part of doc you're pointing to?
2. The original statement is correct, ref. https://docs.google.com/document/d/1Qbgkhl44AYOmkSp6JfT_iDL9IVYRxRFE/edit

docs/cep/query-guide/stream.md
jesusgonzalez12: ok, thanks.

docs/cep/query-guide/stream.md
jesusgonzalez12: changing for "Syntax". Is that right?

jesusgonzalez12:
Hi @StoyanTc 

1) Here https://macrometa.com/docs/cep/query-guide/stream#sink on "Map Payload" subsection, last paragraph.
2) OK, good to know, thanks.

diana-macrometa:
What is the status of this, @jesusgonzalez12 ?

jesusgonzalez12:
Thanks for keeping an eye @diana-macrometa.

The only thing pending is my question about the added paragraph I've been talking about, my point number one question in a previous comment.

@StoyanTc, I have provided the location of the paragraph. Would be great to have more context or an example.

StoyanTc:
> Thanks for keeping an eye @diana-macrometa.
> 
> The only thing pending is my question about the added paragraph I've been talking about, my point number one question in a previous comment.
> 
> @StoyanTc, I have provided the location of the paragraph. Would be great to have more context or an example.

@jesusgonzalez12 - 

- If the key is `$abc='some value'` the output will be `{"$abc": "some value"}`. This is just a constant.
- If the key is `a.b.c="some value", the output will be nested json like `{"a" : {"b":{"c":"some value"}}}`

jesusgonzalez12:
Thanks @StoyanTc.

@diana-macrometa, adding Stoyan's example can improve de doc. What do you think?

diana-macrometa (approved):
Looks good!

PR #419, opened 2022-12-01
Update redis docs

diana-macrometa:
Added documentation for Redis Mode collections indexes and settings.


docs/collections/redis-mode/redis-mode-indexes.md
```diff
@@ -1,26 +1,19 @@
 ---
-title: Document Store Indexes
-sidebar_position: 50
+title: Redis Mode Indexes
```
martin-macrometa, line 2: Is Indexes correct english or should we use Indices?

docs/collections/redis-mode/redis-mode-indexes.md
```diff
@@ -1,26 +1,19 @@
...
 
-You can create the following types of index for a document collection:
-
-- [Geo Index](../indexing/geo-indexes.md)
-- [Fulltext Index](../indexing/fulltext-indexes.md)
-- [Persistent Index](../indexing/persistent-indexes.md)
-- [TTL Index](../indexing/ttl-indexes.md)
-
-Refer to each index type for more information about the index and for instructions on how to create and use it.
+Redis Mode collections have several indexes created automatically. You cannot create additional indexes in Macrometa for these collections.
```
martin-macrometa: "You cannot create additional indexes" - technically makes sense but for the user a more useful message would be "You don't need to create additional indexes"

docs/collections/redis-mode/redis-mode-indexes.md
diana-macrometa: Both versions are correct, but "indexes" is simpler to understand and is the version we most often use in technical writing.

docs/collections/redis-mode/redis-mode-indexes.md
diana-macrometa: As a user, I would be surprised by that statement and would want more explanation. WHY do you say the user will not need additional indexes?

docs/collections/redis-mode/redis-mode-indexes.md
dustinlarimer: @diana-macrometa @martin-macrometa +1, we also use "indexes" in the GUI, so I would prefer we stick with that.

docs/collections/redis-mode/redis-mode-indexes.md
dustinlarimer: IIRC, I think the message here is that we create all of the required indexes up front, so there's no further optimization needed by the user. If that's correct, we should just say exactly that.

diana-macrometa: My inner user goes, "But what about MY use case, which is special?" I'm wondering how we can guarantee that all indexes that might be required are there.

martin-macrometa: The Redis API supports a fixed set of commands. Most of those operate on a single entry and don't need any special indexing (the purpose of indexes is basically to speed up operations that somehow involve multiple entries). We have taken care to optimize the commands operating on multiple entries (or we are in the process of optimizing those).

diana-macrometa: Perfecto. How's this text?

PR #427, opened 2022-12-02
Update Crypto Trading Bot and Fleet Management app pages

nobuyoshi-macrometa:
### Notes
These changes are related to the migration of the demo apps to the new Play platform.
New Tenants (email), Fabric names and Region names have been created. We updated the Sample Apps pages with these new values.


diana-macrometa (approved):
Looks good, only suggestion is I would crop the image and shrink it so that it is no greater than 1000 pixels wide. Right now, it is way too big.

@dustinlarimer might have more specific advice on that.

nobuyoshi-macrometa:
> Looks good, only suggestion is I would crop the image and shrink it so that it is no greater than 1000 pixels wide. Right now, it is way too big.
> 
> @dustinlarimer might have more specific advice on that.

Ok. I'll do that.

PR #449, opened 2022-12-08
Reconcile Apache Kafka docs

diana-macrometa:
@StoyanTc, this one has some issues and needs a thorough review. There are several discrepancies between the Functions and Extensions versions of the Apache Kafka docs.

- Apache Kafka in Extensions is labeled beta. Is that because of the Kafka Replay Request docs?
- The Kafka Replay Request for source and sink is only in the Extensions version. Should it be transferred to the Functions folder or discarded?
- The Source has `enable.async.commit` in the parameter list in the Extensions version, but not the Functions version. Please reconcile.
- Extensions Source example has three examples, the Functions source only has two. Should we transfer that third example or discard it?


docs/cep/query-guide/functions/sink/kafka.md
```diff
@@ -45,4 +46,4 @@ This Kafka sink configuration publishes to 0th partition of the topic named `top
     insert into BarStream
```
bdzekic1, line 46: `@info(name = 'query1)` makes more sense to go above the query than above sink creation.
Also, map.type should be `json` instead of `xml`, since we don't currently support `xml`.
This should be changed in both examples.

docs/cep/query-guide/functions/source/kafka.md
```diff
@@ -8,25 +8,24 @@ with a partition for a Kafka cluster. The events received can be in the
...
+| optional.configuration | This parameter contains all the other possible configurations that the consumer is created with. e.g., `ssl.keystore.type:JKS,batch.size:200`.                               | null          | STRING              | Yes      | No      |
 
 ## Example 1
```
bdzekic1, line 30: Similar changes like in the sink->kafka, should be made in this file.
@info(name = 'query1) makes more sense to go above the query than above sink creation.
Also, map.type should be json instead of xml, since we don't currently support xml.
This should be changed in both examples.

bdzekic1 (changes requested):

StoyanTc (approved):

docs/cep/query-guide/functions/sink/kafka.md
diana-macrometa: Please either make a commit on this branch PR or make the changes in suggestions.

docs/cep/query-guide/functions/source/kafka.md
diana-macrometa: Please either make a commit on this branch PR or make the changes in suggestions. To make a suggestion, click the icon with a plus and minus in the comment editor.

diana-macrometa: @bdzekic1, I do not understand what changes you are asking for. Please either make a commit or make a suggestion as I described.

bdzekic1: I have, you should be able to see changes right now @diana-macrometa

diana-macrometa: I meant a commit on this PR. It's fine, I'll reconcile the conflicts once the other PR is merged.

In future, please feel free to put a commit on the PR branch (move-apache-beta in this case). Then I won't have to reconcile conflicts later.

bdzekic1: Understood. Sorry for the misunderstanding. Will keep this in mind for future PRs. There will probably be another PR that I did outside of this branch if you haven't received it already.

docs/cep/query-guide/functions/sink/kafka.md
diana-macrometa: @StoyanTc, can you help me understand what he means here? Or just make the changes yourself, if that is faster?

bdzekic1: @diana-macrometa changes have already been made in separate PR, which you have merged earlier, so this comment can be dismissed

PR #459, opened 2022-12-12
Update query.md

bdzekic1:
Updated syntax and renamed `Stream App` to `Stream Worker`

PR #460, opened 2022-12-12
Update query.md

bdzekic1:
Changed syntax and renamed `Stream App` to `Stream Worker`


diana-macrometa:
@bdzekic1, this seems very similar to your other Update query.md branches.

bdzekic1:
It is similar since they are part of the same doc, but I wanted to divide 1PR per section. Still, I think I made similar changes, so can we merge another Update query.md to master, I think it's 459, and then update this branch so that I can re-write any conflicting changes, and then we can merge this one also? @diana-macrometa

docs/cep/query-guide/query.md
```diff
@@ -866,7 +866,7 @@ Square brackets can be used to indicate the event index where `1` can be used as
 
 **Example**
 
-The following Stream App calculates the temperature difference between two regulator events.
+The following Stream Worker calculates the temperature difference between two regulator events.
```
diana-macrometa: Same edits as the last one. Lowercase stream worker.

docs/cep/query-guide/query.md
```diff
@@ -955,15 +955,15 @@ select e1.roomNo as roomNo
 from e1=RegulatorStateChangeStream[action == 'start'] -> not TempStream[e1.roomNo == roomNo and temp < 12] and e2=RegulatorStateChangeStream[action == 'off'];
 ```
 
-This Stream Application generates an alert if the temperature does not reduce to 12 degrees within 5 minutes of switching on the regulator.  
+This Stream Worker generates an alert if the temperature does not reduce to 12 degrees within 5 minutes of switching on the regulator.
```
diana-macrometa: Write out numbers zero through ten as words.

diana-macrometa (changes requested):

docs/cep/query-guide/query.md
bdzekic1: @diana-macrometa could you explain to me where would you like me to do that since I'm not sure I understand the task?

diana-macrometa: Replace "5" with "five."

diana-macrometa (approved):

diana-macrometa:
Please resolve the conflicts.

diana-macrometa:
#459 was already merged.

bdzekic1:
I think everything is consistent. This should be merged to update code in line 988 from `'5 min'` to `5 min`..changes have already been made, just needs to be merged.

PR #473, opened 2022-12-14
Update xml.md

bdzekic1:
I'm not sure what is the best way to delete this page is, so I'll just remove the whole content, and then it can be deleted from your side properly. Since we currently don't support `map.type=xml`, this page will create confusion among customers, so removing it would be the best option ATM. This has been confirmed in a thread in the slack channel by Stoyan and advised by James to be removed.


diana-macrometa:
Literally, delete the file. If you have cloned the repo (which you should so that you can do local builds and check your work), then you delete the file like any other in your finder.

If you are only working on the web view, then click the three dots in the upper right of the file window and click Delete File in the drop-down list.

PR #478, opened 2022-12-14
Update http-service.md

bdzekic1:
Removed any mention of `xml` and replaced example with more recent one


diana-macrometa (approved):
LGTM!

PR #501, opened 2022-12-20
SUP-141 StreamWorker Named Aggregation Audit

koshy-thomas:
1. Adding attributes (store.type='database' and store.replication.type='global') as without these by default local aggregation collections are created but since local collections are disable it will fail.
2. Changing the query order and correcting as per QL version 2. Like some places FROM statement is after GROUPBY, INSERT statement is after SELECT, etc.
3. Removing AGG_TIMESTAMP from Query only as the query tries to create the Aggregations but AGG_TIMESTAMP is present inside the Aggregations that is after it is created should it be retrieved.
4. Changing StockStream examples to TradeStream examples to stay consistent with an example.


docs/cep/query-guide/named-aggregation.md
```diff
@@ -32,11 +32,11 @@ The above syntax includes the following:
 |Item                          |Description
 ---------------                |---------
 |`store`                      |This annotation is used to refer to the data store where the calculated aggregate results are stored. This annotation is optional. When no annotation is provided, the data is stored in the `in-memory` store.
-|`purge`                      |This annotation is used to configure purging in aggregation granularities. If this annotation is not provided, the default purging mentioned above is applied. If you want to disable automatic data purging, you can use this annotation as follows:`purge.enable='false'`/You should disable data purging if the aggregation query in included in the Stream application for read-only purposes.
+|`purge`                      |This annotation is used to configure purging in aggregation granularities. If this annotation is not provided, the default purging mentioned above is applied. If you want to disable automatic data purging, you can use this annotation as follows:`purge.enable='false'`/You should disable data purging if the aggregation query in included in the Stream worker for read-only purposes.
```
diana-macrometa:
```suggestion
|`purge`                      |This annotation is used to configure purging in aggregation granularities. If this annotation is not provided, the default purging mentioned above is applied. If you want to disable automatic data purging, you can use this annotation as follows:`purge.enable='false'` You should disable data purging if the aggregation query in included in the stream worker for read-only purposes.
```

docs/cep/query-guide/named-aggregation.md
```diff
@@ -32,11 +32,11 @@ The above syntax includes the following:
...
 |`purge.retention.Period`            |This annotation is used to specify the length of time the data needs to be retained when carrying out data purging. If this annotation is not provided, the default retention period is applied.
 |`<aggregator name>`           |This specifies a unique name for the aggregation so that it can be referred when accessing aggregate results.
 |`<input stream>`              |The stream that feeds the aggregation. **Note! this stream should be already defined.**
-|`group by <attribute name>`   |The `group by` clause is optional. If it is included in a Stream application, aggregate values  are calculated per each `group by` attribute. If it is not used, all the events are aggregated together.
+|`group by <attribute name>`   |The `group by` clause is optional. If it is included in a Stream worker, aggregate values  are calculated per each `group by` attribute. If it is not used, all the events are aggregated together.
```
diana-macrometa:
```suggestion
|`group by <attribute name>`   |The `group by` clause is optional. If it is included in a stream worker, aggregate values  are calculated per each `group by` attribute. If it is not used, all the events are aggregated together.
```

docs/cep/query-guide/named-aggregation.md
```diff
@@ -61,38 +61,37 @@ the aggregation input stream should only feed events to one aggregation definiti
 
 **Example**
 
-This Stream Application defines an aggregation named `TradeAggregation` to calculate the average and sum for the `price` attribute of events arriving at the `TradeStream` stream. These aggregates are calculated per every time granularity in the second-year range.
+This Stream worker defines an aggregation named `TradeAggregation` to calculate the average and sum for the `price` attribute of events arriving at the `TradeStream` stream. These aggregates are calculated per every time granularity in the second-year range.
```
diana-macrometa:
```suggestion
This stream worker defines an aggregation named `TradeAggregation` to calculate the average and sum for the `price` attribute of events arriving at the `TradeStream` stream. These aggregates are calculated per every time granularity in the second-year range.
```

diana-macrometa: The second sentence of this makes no sense.

docs/cep/query-guide/named-aggregation.md
```diff
@@ -61,38 +61,37 @@ the aggregation input stream should only feed events to one aggregation definiti
...
   from TradeStream
-  select symbol, avg(price) as avgPrice, sum(price) as total
     group by symbol
     aggregate by timestamp every sec ... year;
 ```
 
 ### Distributed Aggregation
 
-Distributed Aggregation allows you to partially process aggregations in different shards. This allows Stream app in one shard to be responsible only for processing a part of the aggregation.
+Distributed Aggregation allows you to partially process aggregations in different shards. This allows Stream worker in one shard to be responsible only for processing a part of the aggregation.
```
diana-macrometa:
```suggestion
Distributed aggregation allows you to partially process aggregations in different shards. This allows stream worker in one shard to be responsible only for processing a part of the aggregation.
```

docs/cep/query-guide/named-aggregation.md
```diff
@@ -61,38 +61,37 @@ the aggregation input stream should only feed events to one aggregation definiti
...
-CREATE AGGREGATION <aggregator name> WITH (PartitionById.enable='false')
+CREATE AGGREGATION <aggregator name> WITH (store.type='database', store.replication.type='global', PartitionById.enable='false')
 select <attribute name>, <aggregate function>(<attribute name>) as <attribute name>, ...
+from <input stream>
     group by <attribute name>
     aggregate by <timestamp attribute> every <time periods> ;
-from <input stream>
 ```
 
 Following table includes the `annotation` to be used to enable distributed aggregation,
```
diana-macrometa:
```suggestion
Following table includes the `annotation` to be used to enable distributed aggregation:
```

diana-macrometa (changes requested):

docs/cep/query-guide/named-aggregation.md
koshy-thomas: It means that the aggregations will be calculated per whatever time granularity you have specified (whether "sec" or "min" or "hour" or "day" or "month" or "year") These six values are acceptable and that is what it means by sec-year range it seems.

diana-macrometa (approved):

PR #515, opened 2022-12-22
Update Max Dispatch Throttling Rate

Luba89:
I believe this is in bytes; the scale (metered) tier has a limit of 512000(512 KB) and so this should be 4 KB


Luba89:
@elof can you confirm this

diana-macrometa (approved):

elof:
@Luba89 confirmed

PR #541, opened 2023-01-04
Updated the Javascript full demo file

ghost:
Replaced the Full Demo File for Javascript and added some clarity throughout


docs/streams/getting-started-streams.md
```diff
@@ -19,6 +19,8 @@ This page guides you through creating a stream, publishing messages to it, and s
 
 <Steps />
 
+**Note:** The code snippets in each step aren't the same as the code in the full demo file at the end.
```
diana-macrometa, line 22:
```suggestion
:::note
 The code snippets in each step aren't the same as the code in the full demo file at the end.
 :::
```

diana-macrometa: This is the correct format for notes, cautions, tips, and warnings.

docs/streams/getting-started-streams.md
```diff
@@ -109,9 +111,11 @@ async function streams() {
     try{
       await console.log("Creating local stream...");
       const stream_local = await client.createStream("testStream-local", true);
-
-      await console.log("Creating global stream...");
+      console.log('Local Stream:', stream_local.result["stream-id"]);
+      
+      await console.log("\nCreating global stream...");
```
diana-macrometa, line 116: Why is the /n necessary? We do not have that in any of the other console.log lines I have seen.

docs/streams/getting-started-streams.md
```diff
@@ -323,101 +327,148 @@ else:
...
 
-    } catch(e) {
-      await console.log("Streams could not be created because "+ e);
+  producer.on("open", () => {
+    const input = readline.createInterface({
+      input: process.stdin,
+      output: process.stdout
+    });
+
+    // To repeatedly ask the user for message to be published to the stream. User can always exit by typing 0
```
diana-macrometa:
```suggestion
    // Repeatedly ask the user for message to be published to the stream. User can always exit by typing 0
```

docs/streams/getting-started-streams.md
```diff
@@ -323,101 +327,148 @@ else:
...
-  }
+async function receiveData () {
+  console.log("\n ------- Receive Messages  ------");
+  const consumer = await client.createStreamReader(
+    stream,
+    "test-subscription-1",
+    is_local
+  );
+  
+  // To close consumer connection when user types 0
```
diana-macrometa:
```suggestion
  // Close consumer connection when user types 0
```

diana-macrometa: Yeah, this sort of thing would not make as much sense in the step-by-step. I think it works well here, though.

docs/streams/getting-started-streams.md
diana-macrometa: Let's add a bit about why, though. Something like:

The step-by-step code snippets are designed to run iteratively, so they can build on each other and work even though the app is incomplete.

The full demo code is written as a complete app, with defined functions that are all called at the end, so more like a proper program.

diana-macrometa: If you had time to turn the functions into steps, that would be cool. I understand that you probably do not. Maybe write a Jira ticket describing it so it doesn't get lost?

diana-macrometa (changes requested):

docs/streams/getting-started-streams.md
koshy-thomas: @diana-macrometa I am afraid If we make this full demo file into step by step in getting-started-streams.md then it might be a bad experience for the user as the basic requirement for the user to send message to stream is much more simpler but we add additional code so that the user can have better experience using our full demo files, like adding user input logic, loops, readline interface, etc,

So if we break this into step by step the code becomes a bit more complicated for user to understand what exact code snippet will he need to integrate it with his own code.

diana-macrometa: Understood

docs/streams/getting-started-streams.md
koshy-thomas: exactly! that's why I am not sure if we should break this exactly into step by step or not..

Or there is another option.. we can keep this good full demo file in just https://macrometa.com/docs/streams/stream-examples/pub-sub-streams

And we can keep just simple hardcoded message "Hello World" in Getting Started page( https://macrometa.com/docs/streams/getting-started-streams)?

koshy-thomas: cc @dallas-macrometa

docs/streams/getting-started-streams.md
ghost: It's for readability since it puts the next step on a new line. Without it, there would be 4 output lines in a row vs separated based on the type of stream that was created

docs/streams/getting-started-streams.md
diana-macrometa: I think the full demo file is better code and will be more useful for regular developers, but I think the step-by-step explanations hold value for less advanced users.

diana-macrometa (approved):

docs/streams/getting-started-streams.md
koshy-thomas: yup!

diana-macrometa: We just need to make sure we explain what is going on so that users do not get confused. I suspect that the more advanced folks will skip ahead to the full demo anyway, because that's what devs seem to do.

koshy-thomas (approved):

koshy-thomas:
@diana-macrometa just confirming once, are we good to merge this?

diana-macrometa:
Provided @dallas-macrometa is happy with it, yes.

ghost:
All good with me @diana-macrometa and @koshy-thomas!

diana-macrometa:
In that case, @dallas-macrometa, merge when ready. You have permissions.

PR #543, opened 2023-01-05
DOC-532: Split basics examples

diana-macrometa:
- Reviewed content of the Basics topic. Turned out, it was four basic examples.
- Split examples into separate topics and moved them to the Examples folder.
- Minor format and grammar editing.


diana-macrometa:
@dustinlarimer and @brandonLyonMM, any idea what temp-test is and why it failed?

docs/cep/examples/source-and-sink-example.md
```diff
@@ -0,0 +1,59 @@
...
+    "sensorId":"aq-14",
+    "temperature":35.4
+}
+```
+
+To process custom input messages, refer to Sinkmapper in [Functions](../query-guide/functions/index.md).
+
+## Output
+
+After processing, the event arriving at `TemperatureOnlyStream` will be emitted via `c8stream` and `kafka` sinks.
```
StoyanTc: @diana-macrometa - We have not provided a sample with `Kafka` sink. Why we mention it here?

docs/cep/examples/source-and-sink-example.md
```diff
@@ -0,0 +1,59 @@
...
+
+The message is published to `TemperatureOnlyStream` as
+
+```json
+{"temperature":"35.4"}
+```
+
+The `kafka` sink maps the event to a custom JSON message as below and publishes it to the `temperature` topic.
```
StoyanTc: @diana-macrometa - The same as above, why we have `kafka` here?

docs/cep/examples/source-and-sink-example.md
diana-macrometa: Heck if I know. This is how it was before I edited it. All I did was break up the examples, I didn't change the content.

docs/cep/examples/source-and-sink-example.md
diana-macrometa: Same answer as above. Should we just take it out?

docs/cep/examples/source-and-sink-example.md
StoyanTc: yes, let take it out.

docs/cep/examples/source-and-sink-example.md
StoyanTc: Here also, let's take out `kafka`.

StoyanTc (approved):

docs/cep/examples/source-and-sink-example.md
diana-macrometa: I updated the source-and-sink-example. Please take one more look to ensure that it looks right.

docs/cep/examples/source-and-sink-example.md
```diff
@@ -0,0 +1,53 @@
...
+    "sensorId":"aq-14",
+    "temperature":35.4
+}
+```
+
+To process custom input messages, refer to Sinkmapper in [Functions](../query-guide/functions/index.md).
+
+## Output
+
+After processing, the event arriving at `TemperatureOnlyStream` will be emitted via `c8stream` sink.
```
StoyanTc, line 45: @diana-macrometa - I'm wandering whether we should not replace `c8stream` by `stream` or `gdn stream`? As far as I know `c8streams` is only for internal use.

diana-macrometa: Except that it `c8local` and `c8local` are automatically added as a prefix for every stream that is created. I would prefer `gdn stream` or `stream`, because I have been told we should move away from `c8` wherever possible for branding reasons.

Which option is more correct, in your opinion? gdn stream or stream?

diana-macrometa: I decided to go with stream, because we specify Macrometa in source, so that can be construed to include the sink.

PR #544, opened 2023-01-05
DOC-533 - Query guide index

diana-macrometa:
- Reviewed and edited content of the Stream Worker Query Guide index.
- Changed StreamApp to stream worker.
- Changed order of sections to make them lead more clearly from general to specific.
- Moved some content to the Stream Worker intro page.
- Added links to the elements table for sections that exist.

@StoyanTc - Is Stream QL still a thing that we want to talk about? I haven't heard it used in conversation, really.
- Are the syntax examples correct? I thought you needed a SOURCE and SINK to have a valid stream worker. And I don't even know what to make of the high-level syntax example.
- Do we have content somewhere for event, inner stream, or stream/query callback? I'm not familiar with any of those except event, and I could not find anything obvious to link any of these three items to.


diana-macrometa:
@dustinlarimer and @brandonLyonMM, what's with the new test that is failing?

brandonLyonMM:
> @dustinlarimer and @brandonLyonMM, what's with the new test that is failing?

@diana-macrometa Sorry I didn't realize that error would appear in this project. It's a temporary thing you can safely ignore. I was working on a forked branch and different Vercel project to facilitate the upcoming marketing website migration.

docs/cep/query-guide/index.md
```diff
@@ -58,8 +25,12 @@ High level syntax of StreamApp is as follows.
         ;
 ```
 
+### Syntax Example
+
 For example, this stream application with the name `Temperature-Analytics` creates a stream named `TempStream` and a query named `5minAvgQuery`.
 
+Stream workers are named by adding `@app:name('<name>')` annotation on the top of the stream worker spec. When the annotation is not added, Macrometa assigns a random UUID as the name of the stream worker.
+
```
StoyanTc, line 33: @diana-macrometa - The `OutputStream` is internal stream(sink) since does not have an explicit definition.
Let's replace it by `gdn` stream by adding the line below right after the other definition.
```CREATE SINK STREAM OutputStream (roomNo int, avgTemp double);```

diana-macrometa: Done. Is this in the right place?

Btw, you can insert suggestions where you want to by clicking the icon with the plus and minus in the top of the comment editor.

StoyanTc: Yes, that's the right place.
Somehow it didn't give to use `+/-` previous time.

StoyanTc (approved):

PR #546, opened 2023-01-06
Updating JS/Python code of PubSub example to a better version from Getting-Started page

koshy-thomas:
Just updating Javascript and Python examples for PubSubStreams with the full demo version files in getting started: https://github.com/Macrometacorp/docs/pull/541/files


diana-macrometa (approved):
Thanks, Koshy!

PR #550, opened 2023-01-06
Enrich Data with a Collection Task

diana-macrometa:
This PR completely reworks the existing [Enrich Data with a Collection](https://macrometa.com/docs/cep/enrich-data/enrich-data-with-collection) topic.

I tested the procedure and tried to include enough explanation for people that want it as well as mechanisms for more advanced users to skip ahead.

I could not figure out how to get the Publish message API to accept multiple messages, so the last two steps are kind of clunky. If anyone knows how to do that, please let me know and I will rewrite that part.

Most of the explanation, particularly the query parts, are from the legacy documentation, so please read them thoroughly.

This is the first PR attempt at reworking this sort of tutorial content into a task, so I am hoping for a lot of good feedback to guide us going forward.


StoyanTc (approved):

diana-macrometa:
@StoyanTc, would "join" be a better verb to use here than "enrich"? Join Data with a Collection?

PR #556, opened 2023-01-09
Update redirects for graphs

dustinlarimer:


src/plugins.js
```diff
@@ -32,6 +32,10 @@ const redirectsPlugin = [
       {
         from: '/troubleshooting',
         to: '/references/troubleshooting'
+      },
+      {
+        from: '/collections/graphs',
```
diana-macrometa, line 37: We do not want this behavior. `collections/graph-edge` exists. Right now, it is unclear whether a user will want to look at Graph Edge Collections or Graphs (general functionality), because they were both mixed together.

diana-macrometa (changes requested):

src/plugins.js
dustinlarimer: @diana-macrometa this won't affect `collections/graph-edge`. The original `collections/graphs` URL 404s now, so this just points that exact path to a new home, but `graph-edge` pages will not be affected.

diana-macrometa: okay

diana-macrometa (approved):

PR #561, opened 2023-01-10
Doc-537: Review Partition

diana-macrometa:
- Broke partion.md into separate parts, placed in folder.
- Minor format and grammar edits.
- Added links to new topics to Query Guide index.

@StoyanTc - An example seems to be missing from purge-partition.md. Can you fill it in?


docs/cep/query-guide/partition/purge-partition.md
```diff
@@ -0,0 +1,53 @@
...
+    insert into <output stream>
+    select <attribute name>, <attribute name>, ...
+    from <input stream> ...
+end;
+```
+
+When using purge with an aggregation, use a `WITH()` property instead. For example:
+
+```
+
```
diana-macrometa: @StoyanTc, can you fill in the missing example? Or should I just delete lines 26 through 30?

StoyanTc: @diana-macrometa - This document is only for partitions why we should add example with `aggregation`?

diana-macrometa: This is how the document was when I started editing it, I did not add anything, just am wondering what to do with this empty example. Should I just delete it?

StoyanTc: Yes, I think we need to remove the line containing `aggregation`.  This seems to be a `copy/paste` from somewhere else.

diana-macrometa: Ready for review.

StoyanTc (approved):

PR #574, opened 2023-01-12
Canonical updates

brandonLyonMM:
From `https://macrometa.com/docs` TO `https://www.macrometa.com/docs`

### DIFF
--- a/docs/account-management/api-keys/create-api-keys.md
+++ b/docs/account-management/api-keys/create-api-keys.md
@@ -117,7 +117,7 @@ else:
 </TabItem>
 <TabItem value="RA" label="REST API">
 
-Use our interactive API Reference with code generation in 18 programming languages to [Create an API Key](https://macrometa.com/docs/api#/operations/CreateApiKey).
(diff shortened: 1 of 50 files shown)


docs/account-management/api-keys/create-api-keys.md
```diff
@@ -117,7 +117,7 @@ else:
 </TabItem>
 <TabItem value="RA" label="REST API">
 
-Use our interactive API Reference with code generation in 18 programming languages to [Create an API Key](https://macrometa.com/docs/api#/operations/CreateApiKey).
+Use our interactive API Reference with code generation in 18 programming languages to [Create an API Key](https://www.macrometa.com/docs/api#/operations/CreateApiKey).
```
dustinlarimer, line 120: @brandonLyonMM wondering if we should just make these relative links.. wdyt?

brandonLyonMM: That works for me. I didn't know if there was a reason they weren't relative.

diana-macrometa: Because they are linking to the API reference, which I did not think we could make relative?

PR #585, opened 2023-01-17
DOC-586: CREATE SINK

diana-macrometa:
- Completely reworked the Sink section, now called CREATE SINK.
- Combined content from a couple sources to write a generic CREATE SINK intro.
- Moved sink types into a subfolder.
- Removed content from stream-sink.md to create topis about distributed sinks, sink mapping, and error handling at sink.


docs/cep/sink/sink-types/stream-sink.md
```diff
@@ -0,0 +1,32 @@
...
+
+To configure a stream to publish events via a sink, add the sink configuration to a stream definition by adding the `sink.type` annotation with the required parameter values.
+
+The sink syntax is as follows:
+
+```sql
+   CREATE SINK SinkName WITH (type="stream", stream="STRING", replication.type="STRING", map.type='type') (strings);
+```
+
+## Query Parameters
```
diana-macrometa, line 21: @StoyanTc - This list of query parameters does not seem to match the syntax shown above. Is it correct?

StoyanTc: @diana-macrometa - You're right. Only the `stream source` has `stream.list`.

diana-macrometa: So how should I change it? If you want, you can make the change as a suggestion or a commit.

diana-macrometa: Actually, I think I might have found the right one.

StoyanTc: Yes, that's correct. You was faster than me.

diana-macrometa: You gave me the clue, I copy/pasted the wrong table. So based on what you said, I was able to find the right table.

jamesm-macrometa (approved):

diana-macrometa:
@StoyanTc, is this PR good to go, or do you need to review it some more?

StoyanTc (approved):

PR #602, opened 2023-01-26
DOC-575: Ad hoc queries

diana-macrometa:
This PR creates a new section for Ad Hoc Queries.

- Takes previous content from store.md, edits it and puts it in the Ad Hoc Queries intro and the examples page.
- Creates a new topic showing users how to perform ad hoc queries.


docs/cep/ad-hoc-queries/index.md
```diff
@@ -0,0 +1,56 @@
...
+```
+
+In order to execute ad hoc queries, the stream worker you are using should have a sink or defined that contains the table, window, or aggregator to be queried.
+
+## Next Steps
+
+- [Perform ad hoc queries](perform-ad-hoc-queries.md).
+- [Look at ad hoc query examples](ad-hoc-examples.md).
+
+## Event Playback
```
diana-macrometa: @StoyanTc, this Event Playback bit was in the original store.md topic. It seems out of place, but I have no idea what to do with it or if it is even valid. Any suggestions?

StoyanTc: @diana-macrometa - We should take this out. I guess there is another place where all such specific annotations are explained?

StoyanTc (approved):

docs/cep/ad-hoc-queries/index.md
diana-macrometa: Not that I am aware of.

diana-macrometa:
I moved Event Playback into a separate topic. We'll revisit it later, perhaps group it with other annotations if we find them.

PR #603, opened 2023-01-26
Update index.md

diana-macrometa:
- Corrected some format errors

@StoyanTc, so we need to adjust this topic? The way it reads, only streams can be partitioned. If that is correct, then we are fine. If it is not, then we need to adjust the content.


diana-macrometa:
@StoyanTc, will you be able to get to this PR this week?

StoyanTc (approved):

PR #606, opened 2023-01-26
Table and window example

diana-macrometa:
This PR removes the example from the Create Stream Workers tutorial and slightly improves the explanation language.

@StoyanTc and @koshy-thomas, please take a look at the code in the example. When I try to validate, the validation is complaining about the CREATE TABLE line. It looks exactly like other syntax and code examples, so I don't know what's up here.


koshy-thomas:
@diana-macrometa which federation are you using since I didn't get any validation errors when I validate that SW. Let's connect on slack to check this issue.

StoyanTc (approved):
@diana-macrometa - 
1. The validation of the stream worker passed for me too.
2. I did a minor change into the file changing `group by` and `having` to capital letters.

diana-macrometa:
I checked the validation again (in Playground both times, btw), and it is working fine. No idea why it gave me an error yesterday.

PR #610, opened 2023-01-27
Update realtime-updates.md

jamesm-macrometa:
The purpose of this PR is to revise the "Realtime DB Updates Example" topic and add missing context information.


jamesm-macrometa:
@diana-macrometa I'm ready for you to take a look at this, but I have some outstanding questions in Slack about how to view the "callback" notifications we subscribed to.

docs/streams/stream-examples/realtime-updates.md
```diff
@@ -1,33 +1,40 @@
 ---
 sidebar_position: 140
-title: Realtime DB Updates Example
+title: Real-Time DB Updates
```
diana-macrometa: Removing "Example" from the title breaks the pattern of titles in this section. Please revert this change or we can discuss removing "Example" from the other titles.

docs/streams/stream-examples/realtime-updates.md
```diff
@@ -74,8 +105,172 @@ async function createCollection () {
...
+
+## Full Demo File
+
+<Tabs groupId="operating-systems">
+<TabItem value="js" label="Javascript">
+
+```js
+const jsc8 = require("jsc8");
+
+const client = new jsc8({url: "https://play.paas.macrometa.io", apiKey: "XXXX", fabricName: '_system'});
```
diana-macrometa: I just realized that we should add a comment everywhere we use the Play link to instruct users to update this if they are not using Play. Please add appropriate comments in the partial file or create a Jira so we do it later (but soon). We'll also need to review this in places where we don't single source.

docs/streams/stream-examples/realtime-updates.md
```diff
@@ -74,8 +105,172 @@ async function createCollection () {
...
+
+// Variables
+const collectionName = "ddos";
+let listener;
+const data = [
+  { ip: "10.1.1.1", action: "block", rule: "blocklistA" },
```
diana-macrometa: Avoid "blacklist" and "whitelist" in all cases. Use blocklist and allowlist instread.

docs/streams/stream-examples/realtime-updates.md
```diff
@@ -102,12 +297,11 @@ import threading
 import pprint
```
diana-macrometa, line 204: It looks like this list of imports does not match what is imported in the step-by-step code. Some are missing.

docs/streams/stream-examples/realtime-updates.md
```diff
@@ -42,17 +49,41 @@ const data = [
...
+# Variables - Data
+data = [
+    {"ip": "10.1.1.1", "action": "block", "rule": "blacklistA"},
+    {"ip": "20.1.1.2", "action": "block", "rule": "blacklistA"},
+    {"ip": "30.1.1.3", "action": "block", "rule": "blacklistB"},
+    {"ip": "40.1.1.4", "action": "block", "rule": "blacklistA"},
+    {"ip": "50.1.1.5", "action": "block", "rule": "blacklistB"},
+]
 
+pp = pprint.PrettyPrinter(indent=4)
```
diana-macrometa: I got this error after adding the second step. Please be sure to thoroughly test your documentation, both the full demo and the step-by-step versions.

```
(base) dianapayton@Dianas-MacBook-Pro python % python real-time.py
--- Connecting to GDN
Traceback (most recent call last):
  File "/Users/dianapayton/Documents/python/real-time.py", line 33, in <module>
    pp = pprint.PrettyPrinter(indent=4)
NameError: name 'pprint' is not defined
```

docs/streams/stream-examples/realtime-updates.md
```diff
@@ -42,17 +49,41 @@ const data = [
...
+</TabItem>
+</Tabs>
+
+
+### Step 3. Create Collection
+
+Create a collection called `ddos` to which we will subscribe. If a collection by that name already exists, the existing collection is used instead.
+
+<Tabs groupId="operating-systems">
+<TabItem value="js" label="Javascript">
```
diana-macrometa: JavaScript, not Javascript, in all cases.

docs/streams/stream-examples/realtime-updates.md
```diff
@@ -1,33 +1,40 @@
...
+<ConnectToGDN />
+
+### Step 2. Set Variables
+
+Set the variables required to run the code.
 
 <Tabs groupId="operating-systems">
 <TabItem value="js" label="Javascript">
 
 ```js
```
diana-macrometa, line 59: Please work with @koshy-thomas to get some log lines into the JavaScript steps. At least for the first three steps, there is no user response, which is unsatisfying and worrisome to a user going through this step by step.

Be sure to make the same changes to the full demo if appropriate.

docs/streams/stream-examples/realtime-updates.md
```diff
@@ -74,8 +105,172 @@ async function createCollection () {
...
+
+<Tabs groupId="operating-systems">
+<TabItem value="js" label="Javascript">
+
+```js
+async function deleteData () {
+  console.log("\n4. Delete data");
+  await client.deleteCollection(collectionName);
+}
+(async function () {
```
diana-macrometa: This is the part where the function actually runs. I strongly suspect based on other examples I've worked on (double check with @koshy-thomas or test it yourself) that we need to have the run commands in this function up with each of the steps.

As it is, the user gets no feedback when doing the JavaScript step by step.

diana-macrometa (changes requested):
Testing documentation is repetitive, but it is necessary. It is also how you build expertise in this domain.

As written, the user will not be successful if they try to run this example using the incremental method. I see that you wrote the individual steps yourself, which is good, but that means you must test them and make sure you do not submit the PR for review until you are sure that they work.

diana-macrometa:
This is what it looks like when you run the steps of the JS step by step. No feedback until the fifth step.
image

docs/streams/stream-examples/realtime-updates.md
jamesm-macrometa: Good catch, looks like I accidentally deleted a couple things when I messed with the formatting.  Lesson learned - test again after messing with the format!

docs/streams/stream-examples/realtime-updates.md
jamesm-macrometa: @diana-macrometa Is there a way to add comments to the partial file so that the comments don't display in-text?

diana-macrometa: Nope. Markdown does not support comments. 

In this case, we would want the comments to display to readers, though. Same as the `apikey` comments, just something like "If you aren't using Play, then enter your (whatever we call the address)."

jamesm-macrometa: > Nope. Markdown does not support comments.
> 
> In this case, we would want the comments to display to readers, though. Same as the `apikey` comments, just something like "If you aren't using Play, then enter your (whatever we call the address)."

Got it, sounds good!

jamesm-macrometa:
@koshy-thomas Can you take a look at the latest version of the example just in case I overlooked something?

Also, I noticed the JS version of this example fails if the `ddos` collection already exists, whereas the Python version simply uses the existing collection.  How doable would it be to add this same functionality to the JS sample?

koshy-thomas:
> @koshy-thomas Can you take a look at the latest version of the example just in case I overlooked something?
> 
> Also, I noticed the JS version of this example fails if the `ddos` collection already exists, whereas the Python version simply uses the existing collection. How doable would it be to add this same functionality to the JS sample?

@jamesm-macrometa the code looks fine to me, and I don't see the issue you mention in JS too.. even if a collection exists ,the JS code should simply use the existing collection.. do let me know (ping me on slack).. if you still face this issue

docs/streams/stream-examples/realtime-updates.md
```diff
@@ -1,32 +1,201 @@
...
+    # Subscribe to be notified when changes are made to collection.
+
+    def create_callback():
+        def callback_fn(event):
+            pp.pprint(event)
+            return
+
+        client.on_change(COLLECTION_NAME, callback=callback_fn, timeout=15)
+
+    # Step3: Subscribe to receive documents in realtime (PUSH model)
```
koshy-thomas: We don't receive documents exactly.. it should be rephrased I guess, we receive real-time updates/changes made to the collection.. it can be deleting a document or updating a document as well.

docs/streams/stream-examples/realtime-updates.md
```diff
@@ -1,32 +1,201 @@
...
+    print(f"\n3. SUBSCRIBE_COLLECTION: region: {URL},  collection: {COLLECTION_NAME}")
+    rt_thread = threading.Thread(target=create_callback)
+    rt_thread.start()
+    time.sleep(10)
+    print(f"Callback registered for collection: {COLLECTION_NAME}")
+
+    # Step4: Subscribe to receive documents in real-time (PUSH model)
```
koshy-thomas: Here also, maybe we can rephrase it to say something like "Step4: Insert documents into the collection to trigger the subscription/notification" 
cc @diana-macrometa

diana-macrometa:
I'm working through this now. I noticed that the tabs are not consistent, either in order or in name. Please correct that.

image

docs/streams/stream-examples/realtime-updates.md
```diff
@@ -1,32 +1,201 @@
...
+
+</TabItem>
+</Tabs>
+
+
+### Step 3. Add Code
+
+Add code to perform the following actions:
+
+1. Create a collection called `ddos` to which we will subscribe. If a collection by that name already exists, the existing collection is used instead.
```
diana-macrometa: Link to the collection section. Also, specify what collection type.

Reminder: write for "you" not "we". Also, try to get rid of the passive constructions.

docs/streams/stream-examples/realtime-updates.md
```diff
@@ -1,32 +1,201 @@
...
+1. Add `data` to the collection, then subscribe to the collection. In this example, we are adding IP addresses to block.
+1. Delete `data` from the collection. If the collection becomes empty, the collection is deleted too.
```
diana-macrometa:
```suggestion
1. Delete `data` from the collection. If the collection becomes empty, then the collection is deleted too.
```

diana-macrometa: I don't think that's what is happening. The step implies that the collection becoming empty is what triggers the deletion. In the code, though, the script is truncating the collection (emptying it of all records), then deleting it.

@koshy-thomas, am I correct here?

Also, isn't truncating before deleting an unnecessary extra step?

docs/streams/stream-examples/realtime-updates.md
```diff
@@ -1,32 +1,201 @@
...
+```
+
+</TabItem>
+
+<TabItem value="py" label="Python">
+
+```py
+if __name__ == '__main__':
+
+    # Step1: Open connection to GDN. You will be routed to closest region.
```
diana-macrometa: Space between the step and the number, in all cases. Also, I'm not sure we really need to specify step numbers here. I think it could be confusing given the Step 1, Step 2, Step 3 in the side menu.

docs/streams/stream-examples/realtime-updates.md
```diff
@@ -1,32 +1,201 @@
...
-This tutorial is about using Macrometa GDN as a realtime database with local latencies across the globe.
+This tutorial demonstrates how to use Macrometa GDN as a real-time database with local latencies across the globe.
 
-## Prerequisites
+<Prerequisites />
 
-- A Macrometa account with sufficient permissions to create streams.
-- An API key. For more information, refer to [Create API Keys](../../account-management/api-keys/create-api-keys).
-- Appropriate SDK installed. For more information, refer to [Install SDKs](../../sdks/install-sdks.md).
+## Code
```
diana-macrometa: This heading is extremely vague. I suggest something like, "Step-by-Step Instructions"

docs/streams/stream-examples/realtime-updates.md
```diff
@@ -102,33 +267,32 @@ import threading
...
+    {"ip": "30.1.1.3", "action": "block", "rule": "blocklistB"},
+    {"ip": "40.1.1.4", "action": "block", "rule": "blocklistA"},
+    {"ip": "50.1.1.5", "action": "block", "rule": "blocklistB"},
 ]
 
 pp = pprint.PrettyPrinter(indent=4)
 
 if __name__ == '__main__':
 
     # Step1: Open connection to GDN. You will be routed to closest region.
```
diana-macrometa: See earlier comment about steps. Also, make sure the comments in this demo match the ones above.

docs/streams/stream-examples/realtime-updates.md
```diff
@@ -144,25 +308,25 @@ if __name__ == '__main__':
...
     client.insert_document(COLLECTION_NAME, document=data)
 
     # Step5: Wait to close the callback.
     print("\n5. Waiting to close callback")
     rt_thread.join(2)
 
-    # Step6: Delete data.
-    print(f"\n6. DELETE_DATA: region: {GLOBAL_URL}, collection: {COLLECTION_NAME}")
+    # Step 6: Delete data.
+    print(f"\n6. DELETE_DATA: region: {URL}, collection: {COLLECTION_NAME}")
```
diana-macrometa: On the printout, I see play.macrometa.... as the region. That does not seem right; it isn't a region.

docs/streams/stream-examples/realtime-updates.md
```diff
@@ -102,33 +267,32 @@ import threading
...
 ]
 
 pp = pprint.PrettyPrinter(indent=4)
 
 if __name__ == '__main__':
 
     # Step1: Open connection to GDN. You will be routed to closest region.
-    print(f"\n1. CONNECT: federation: {GLOBAL_URL},  user: {EMAIL}")
-    client = C8Client(protocol='https', host=GLOBAL_URL, port=443, apikey="", geofabric=FABRIC)
+    print(f"\n1. CONNECT: federation: {URL}")
```
diana-macrometa: Do not use "federation" in user-facing documents. In most cases, you can use "server."

docs/streams/stream-examples/realtime-updates.md
```diff
@@ -43,53 +212,49 @@ const sleep = (milliseconds) => {
...
+	}
+	await sleep(2000);
+	await insertData();
+
+	async function deleteData () {
+		console.log("\n4. Delete data");
+		await client.deleteCollection(collectionName);
+	}
+	await sleep(10000);
+	// Closing OnChange listener
```
diana-macrometa: Generally, I like to leave an empty line above comments. It helps with readability.

docs/streams/stream-examples/realtime-updates.md
```diff
@@ -43,53 +212,49 @@ const sleep = (milliseconds) => {
...
+	async function createCollection () {
+		console.log("\n1. Log in.");
+		console.log("\n2. Create collection.");
+		try {
+		  console.log(`Creating the collection ${collectionName}...`);
+		  const existsColl = await client.hasCollection(collectionName);
+		  if (existsColl === false) {
+			await client.createCollection(collectionName, { stream: true });
+		  }
+		  // Add an onChange listener for collection
```
diana-macrometa: What is an onChange listener? This comment reads like this is an actual thing, but I've not heard of it. Also, kind of weird that we cite this thing here but not in the Python version.

Also, take note - listeners are one of the things that need to be documented for the Stream Basics tasks.

docs/streams/stream-examples/realtime-updates.md
diana-macrometa:
```suggestion
	
	// Close OnChange listener
```

diana-macrometa: Also, avoid gerunds in favor of active verbs, and don't forget to put periods at the end of sentences.

diana-macrometa (changes requested):
Great stuff. I was able to complete the example without problems, I found it generally easy to understand. I think the JavaScript version might need a few more comments to explain what is going on, but I think this is mostly to the polish phase.

docs/streams/stream-examples/realtime-updates.md
koshy-thomas: @diana-macrometa correct this needs to be rephrased.. we are just truncating/deleting the data first and then deleting the collection (it's not a cause and effect thing)..

And yes, In fact we can just directly delete the collection directly without truncating it.

diana-macrometa: Is there a good reason to do so, or should @jamesm-macrometa just experiment with removing that truncate command?

docs/streams/stream-examples/realtime-updates.md
koshy-thomas: Basically the function name is OnChange and it just listens to updates being done on the collection (via subscribing/reading the collection stream).. You can decide however we want to phrase it to the users and sync both Js and Py files

docs/streams/stream-examples/realtime-updates.md
koshy-thomas: Yeah you can remove the truncate command, it won't make any difference.. unless you want to show the user how he can truncate the collection via the sdks but still this is not right place to show this.. so feel free to remove the truncate command from this example, it won't make a difference

jamesm-macrometa:
@brandonLyonMM Can you help merge the latest `main` build into this branch?  I believe there are some fixed links I haven't picked up which seem to be causing me errors.

jamesm-macrometa:
Thanks again @koshy-thomas !  Can you take one last look and make sure I put the additional/revised comments in the right places?

brandonLyonMM:
> @brandonLyonMM Can you help merge the latest `main` build into this branch? I believe there are some fixed links I haven't picked up which seem to be causing me errors.

It looks like this branch contains the latest changes from master, that should be alright.

There's a "merging is blocked because changes have been requested" thread in this pull request.

Greenshot 2023-02-01 16 39 25

Greenshot 2023-02-01 16 39 28

Greenshot 2023-02-01 16 39 50

Greenshot 2023-02-01 16 40 30

jamesm-macrometa:
Thanks for confirming @brandonLyonMM , I know the merging is blocked by GitHub review, it's the build itself I'm having trouble with.  I'll troubleshoot some more and let you know if I need help.

docs/streams/stream-examples/realtime-updates.md
```diff
@@ -1,33 +1,62 @@
...
+import pprint
+import time
+COLLECTION_NAME = "ddos"
 
-// Constants - DB
-const globalUrl = "https://play.paas.macrometa.io/";
-const email = "your@email.com";
-const password = "password";
-const client = new jsc8(globalUrl);
+# Variables - Data
```
diana-macrometa:
```suggestion
# Variables
```

diana-macrometa: Very minor suggestion - I like to have the comments in the different versions exactly match wherever possible. If you apply this, then also update the full demo.

diana-macrometa (approved):
One super minor suggestion for you to look at, then good to go! Well done!

jamesm-macrometa:
@koshy-thomas Thanks for meeting with me to go over the sample!  I've incorporated your suggested changes.

BTW, don't worry if the comments still don't line up perfectly in the code samples.  It should display properly once rendered on Docusaurus.

PR #618, opened 2023-02-01
upgrade docusaurus to the latest version

elof:


dustinlarimer:
@elof v2.3.0 may have introduced an issue that's still getting worked out. I've rolled back to v2.2.0, which should have the components we need. https://github.com/facebook/docusaurus/issues/8591

PR #620, opened 2023-02-01
Add reference cards to index pages

elof:
This PR adds cards with referenced to content in folders to indexes that do not have reference content


src/theme/MDXComponents.js
```diff
@@ -1,11 +1,13 @@
 import MDXComponents from '@theme-original/MDXComponents';
 
 /* Components */
+import DocCardList from '@theme/DocCardList';
```
dustinlarimer, line 4: @elof I moved this import into `MDXComponents.js` file, so you won't need to use the `import` statement in markdown files anymore.

dustinlarimer (approved):

PR #628, opened 2023-02-02
Remove extra words

diana-macrometa:
I realized that having the descriptor in the title in parentheses not only looked odd, it was superfluous now that the topics are in proper folders. Adding @dustinlarimer to this PR in case he wants to disagree.

- Removed redundant words from titles.
- Some minor format edits.
- No meaningful content changes.


dustinlarimer (approved):
Good idea :+1:

PR #655, opened 2023-02-15
Replace HAVING with WHERE

diana-macrometa:
This PR replaces HAVING with WHERE to make stream worker examples compliant with current code format.

### DIFF
--- a/docs/cep/examples/table-and-window-example.md
+++ b/docs/cep/examples/table-and-window-example.md
@@ -52,5 +52,5 @@ FROM InboundTraffic WINDOW SLIDING_TIME(2 sec) as IT
 JOIN SuspiciousIPTable as SIP
 ON  IT.ip == SIP.blocked_ip
 GROUP BY ip
-HAVING requestCount > 50 ;
+WHERE requestCount > 50 ;
 ```
--- a/docs/cep/examples/table-and-window-example.md
+++ b/docs/cep/examples/table-and-window-example.md
@@ -51,6 +51,6 @@ SELECT ip, count(ip) as requestCount, currentTimeMillis() as incidentTime
 FROM InboundTraffic WINDOW SLIDING_TIME(2 sec) as IT
 JOIN SuspiciousIPTable as SIP
 ON  IT.ip == SIP.blocked_ip
-GROUP BY ip
-WHERE requestCount > 50 ;
+WHERE requestCount > 50
+GROUP BY ip;
 ```
--- a/docs/cep/windows/window-types/tumbling-length.md
+++ b/docs/cep/windows/window-types/tumbling-length.md
@@ -94,7 +94,7 @@ SELECT sensorId,
 FROM TemperatureStream WINDOW TUMBLING_LENGTH(5)
 GROUP BY sensorId
 -- Output events only when `avgTemperature` is greater than or equal to `20.0`.
-HAVING avgTemperature >= 20.0;
+WHERE avgTemperature >= 20.0;
 ```
 
 ### Batch Event Count Aggregation Behavior
--- a/docs/cep/windows/window-types/tumbling-length.md
+++ b/docs/cep/windows/window-types/tumbling-length.md
@@ -92,9 +92,9 @@ SELECT sensorId,
        max(temperature) AS maxTemperature
 -- Aggregate every `5` events in a batch manner.
 FROM TemperatureStream WINDOW TUMBLING_LENGTH(5)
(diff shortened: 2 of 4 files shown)


docs/cep/examples/table-and-window-example.md
```diff
@@ -52,5 +52,5 @@ FROM InboundTraffic WINDOW SLIDING_TIME(2 sec) as IT
 JOIN SuspiciousIPTable as SIP
 ON  IT.ip == SIP.blocked_ip
 GROUP BY ip
-HAVING requestCount > 50 ;
```
StoyanTc, line 55: @diana-macrometa - Incorrect, after `HAVING` always follows `GROUP BY` but not `WHERE`.

docs/cep/windows/window-types/tumbling-length.md
```diff
@@ -94,7 +94,7 @@ SELECT sensorId,
 FROM TemperatureStream WINDOW TUMBLING_LENGTH(5)
 GROUP BY sensorId
 -- Output events only when `avgTemperature` is greater than or equal to `20.0`.
-HAVING avgTemperature >= 20.0;
+WHERE avgTemperature >= 20.0;
 ```
```
StoyanTc: Incorrect, after HAVING always follows GROUP BY but not `WHERE.

docs/queryworkers/c8ql/coming-from-sql.md
```diff
@@ -485,7 +485,7 @@ SELECT YEAR(dateRegister) AS year,
   FROM users
   WHERE active = 1
   GROUP BY year, month
-  HAVING number > 20;
+  WHERE number > 20;
```
StoyanTc: Incorrect, after HAVING always follows GROUP BY but not `WHERE.

docs/cep/examples/table-and-window-example.md
diana-macrometa: According to the Migration Guide, WHERE replaces HAVING. Are you saying that the WHERE should go before GROUP BY?

diana-macrometa: Okay, I think I get it.

StoyanTc: @diana-macrometa - Let me double check it.

StoyanTc: @diana-macrometa - The compiler allows both `where | having` after `group by`. 
However in SQL it's `having` rather than `where`.

diana-macrometa: May be that way in SQL, but Macrometa changed HAVING to WHERE, according to the Migration Guide for QL v2. 

Do we not want to use WHERE? Just reading it, it makes a ton more sense than HAVING, so I would prefer WHERE.

StoyanTc: @diana-macrometa - If I'm a customer and I'm familiar with SQL, construction like this below will look rather strange to me and uncommon.
```
SELECT product_id, count() AS amount
FROM products
WHERE product_id > 1000
GROUP BY product_id
WHERE count() > 10
```
instead of 
```
SELECT product_id, count() AS amount
FROM products
WHERE product_id > 1000
GROUP BY product_id
HAVING count() > 10
```

diana-macrometa: In the latest commit, I moved the WHEREs up above the GROUP BY, which is what I thought you were saying in your earlier comments. Does that satisfy the concern?

StoyanTc: No. `WHERE ... GROUP BY ...` is not equivalent to `GROUP BY ... WHERE | HAVING ...`.
E.g. in my sample above the count() is performed first on a group and then have to be filtered.
So, let's revert back the previous version but keep the `HAVING` after the `GROUP BY`.

diana-macrometa: Should I close the entire PR, or just revert this one piece?

StoyanTc: Yes, makes sense.

diana-macrometa: Which option makes sense? Close the PR? Or revert this one part?

StoyanTc: Please close the PR and create another with the fix only in `docs/cep/windows/window-types/tumbling-length.md`. Let's keep `HAVING` into the other files.

PR #657, opened 2023-02-16
DOC-536 Revise and Break Up Stream Worker Queries

diana-macrometa:
This PR was getting large, so I'm splitting it into two. This is the first part of breaking up query.md.

- Broke out parts of the Query topic into smaller, more manageable topics.
- Updated query syntax so it is current.
- Merged some examples into appropriate topics.
- Integrated window queries content into the Windows section.
- Integrated the functions content that was in query.md into the Functions section.


docs/cep/query-guide/event-types.md
```diff
@@ -0,0 +1,87 @@
...
+Syntax for event types varies on what keywords you are using.
+
+### INSERT INTO
+
+Event type should be defined in between `INSERT` and `INTO` keywords for insert queries as follows.
+
+```sql
+INSERT <event type> INTO <output stream>
+select <attribute name>, <attribute name>, ...
+from <input stream> window <window name>(<parameter>, <parameter>, ... )
```
StoyanTc: The `select` and `from` have remained with small letters.

docs/cep/query-guide/output-rate-limiting.md
```diff
@@ -0,0 +1,82 @@
...
+- Returning events based on time
+
+    Here events are emitted for every predefined time interval. You can also specify whether to emit only the first event, last event, or all events out of the events that arrived during the specified time interval.
+
+    In this example, emits all temperature events every 10 seconds  
+
+    <pre>
+    INSERT INTO LowRateTempStream
+    FROM TempStreamoutput
+    output every 10 sec;    </pre>
```
StoyanTc: `output` should be with capital letters as above and below?

StoyanTc (approved):

PR #659, opened 2023-02-17
Create docs base page

dustinlarimer:
This PR sets up a proper overview page at `/docs` and removes the redirect previously forwarding traffic to `/docs/quickstart`. This update also takes some liberties with the content, and sets up a grid of feature tiles linking off to various parts of the docs site.


docs/index.md
```diff
@@ -0,0 +1,151 @@
+---
+hide_title: true
+sidebar_position: 0
+title: Overview
+---
+
+# Documentation
```
diana-macrometa, line 7: Why not put this text in the title and remove the hide_title?

Also, I suggest something more specific, like Macrometa Documentation as the overview.

docs/index.md
```diff
@@ -0,0 +1,151 @@
...
+    href="/collections/redis-mode/"
+  />
+</grid>
+
+### Edge Compute
+
+Deploy compute runtimes and complex event-driven workloads alongside your data, and within milliseconds of your customers.
+
+<grid cols={3}>
+  <card
```
diana-macrometa, line 93: Queries and query workers are not really separate things. Maybe like to SQL or C8QL here to fill out the cards?

diana-macrometa (approved):
A few suggestions, but this looks great!

docs/index.md
dustinlarimer: @diana-macrometa I'll remove the first tile for now, but we should revisit this. Queries and Query Workers are definitely distinct things with distinct use cases, though we've structured them as one thing here and in the GUI.

docs/index.md
dustinlarimer: @diana-macrometa thanks, "Macrometa" and "Docs" are already safely established by the site navigation scheme. This is an overview page, but since it's also the base page of the docs site it should mirror that placement with a clear page title. I've also set it up this way to avoid name overload, which is a bit unsightly:

image

docs/index.md
diana-macrometa: Depends on which "query" you are talking about, because we have:

- queries you run in the Query Workers tab (or CLI or API or whatever)
- query workers that you save
- ad hoc queries for stream workers
- queries that are part of stream workers
- table queries (stream workers)
- window queries (stream workers)

The documentation for all of which I have been hammering into submission for the past few weeks.

dustinlarimer: @diana-macrometa totally get that, truly some heroic work there!

PR #666, opened 2023-02-17
DOC-612: Unique windows

diana-macrometa:
- Moved window functions in the Unique folder to Window Types
- Minor edits, mostly syntax updates
- Minimal content changes


StoyanTc (approved):
@diana-macrometa - I had to change some of the queries.
Please note `INSERT` is always first, then `SELECT` and finally `FROM`.
The original QL had different syntax.

PR #671, opened 2023-02-21
DOC-536: Break up Query Topic

diana-macrometa:
This is the second part of breaking up query.md.

- Broke topic down into smaller topics.
- Moved Window Queries back into Stream Worker Queries.
- Minor content and format edits throughout.
- Jiras are filed to bring together JOINS and break up Patterns and Sequences in later PRs.


docs/cep/query-guide/having-where.md
```diff
@@ -0,0 +1,57 @@
...
+Query to calculate the average `temp` per `roomNo` for the last 10 minutes, and alerts if the `avgTemp` exceeds 30 degrees.
+
+## Example 2
+
+```sql
+INSERT INTO AlertStream
+SELECT roomNo, avg(temp) AS avgTemp
+FROM TempStream WINDOW SLIDING_TIME(10 min)
+GROUP BY roomNo
+WHERE avgTemp > 30;
```
StoyanTc, line 56: @diana-macrometa - Shall we change this `WHERE` to `HAVING`?

StoyanTc (approved):

docs/cep/query-guide/having-where.md
diana-macrometa: I literally copy/pasted this example from the Migration Guide. Is there a better example we can use for WHERE?

Also, anything we should add to differentiate WHERE and HAVING? The Migration Guide made it sound like they were interchangeable, but I gather that is not the case from your edits.

StoyanTc: Yes, they are interchangeable here but not in SQL. 
Everyone familiar with SQL knows the difference.
```
The difference between the having and where clause in SQL is that the where clause cannot be used with aggregates, but the having clause can.
```

docs/cep/query-guide/having-where.md
```diff
@@ -0,0 +1,59 @@
...
+title: HAVING | WHERE
+---
+
+`HAVING` and `WHERE` provide a way of filtering events based on a specified condition of the query output stream attributes. It accepts any type of condition including a combination of functions and attributes that produces a Boolean result. They allow events to passthrough if the condition results in `true`, and drops if it results in a `false`.
+
+## Purpose
+
+`HAVING` and `WHERE` help to select the events that are relevant for the output based on the attributes those are produced by the `SELECT` clause and omit the ones that are not.
+
+In Macrometa, `HAVING` and `WHERE` are interchangeable. In SQL, the `WHERE` clause cannot be used with aggregates, but the `HAVING` clause can.
```
diana-macrometa: @StoyanTc, how's this?

StoyanTc: @diana-macrometa - Do we need the second sentence on line 12?

diana-macrometa: I thought you wanted it based on the other convo. I can take it out.

diana-macrometa: How's that? Anything else that we should add or remove here?

StoyanTc (approved):

PR #679, opened 2023-02-27
DOC-623: Update query.md with Query Projections

diana-macrometa:
This PR adds information about what projections in stream worker queries can do.

### DIFF
--- a/docs/cep/query-guide/query.md
+++ b/docs/cep/query-guide/query.md
@@ -107,7 +107,7 @@ FROM InputStream;
 
 You can use the `WHERE` or `HAVING` clauses to filter the incoming data stream based on a specific field value.
 
-For example, to select only the customer data where the `age` field is greater than 30, you could use the following query:
+For example, to select only customer data where the `age` field is greater than 30, you could use the following query:
 
 ```sql
 INSERT INTO OutputStream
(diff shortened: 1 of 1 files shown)


docs/cep/query-guide/query.md
```diff
@@ -40,9 +40,132 @@ The following parameters are used to configure a stream definition.
...
+INSERT INTO OutputStream
+SELECT roomNo, temp, 'C' AS scale
+FROM InputStream;
+```
+
+### Filter by Field Value
+
+You can use the `WHERE` or `HAVING` clauses to filter the incoming data stream based on a specific field value.
+
+For example, to select only the customer data where the `age` field is greater than 30, you could use the following query:
```
jamesm-macrometa:
```suggestion
For example, to select only customer data where the `age` field is greater than 30, you could use the following query:
```

docs/cep/query-guide/query.md
```diff
@@ -40,9 +40,132 @@ The following parameters are used to configure a stream definition.
...
+SELECT * 
+FROM InputStream 
+WHERE age > 30;
+```
+
+For more information, refer to [HAVING | WHERE](having-where.md).
+
+### Use Expressions
+
+This uses attributes with mathematical and logical expressions in the precedence order given below, and assigns them to the output attribute using `AS`.
```
jamesm-macrometa: Probably fine, but I stumble a bit reading "This uses attributes" - maybe "Expressions use attributes with..."?

jamesm-macrometa (approved):
Good stuff!  LGTM

StoyanTc (approved):

PR #681, opened 2023-02-28
DOC-625: Create database source

diana-macrometa:
This PR creates a topic to document using a collection as a stream worker source.


docs/cep/table/index.md
```diff
@@ -51,7 +51,7 @@ FROM from_stream … ;
 You can also use general store syntax:
 
 ```sql
-CREATE STORE (GLOBAL|LOCAL)? <table_name> WITH(type="database", propKey=”propVal”, … , PrimaryKey='<attribute_name>', Index='<attribute_name>')(<attribute_name> <attribute_type>, ...);
+CREATE STORE (GLOBAL|LOCAL)? <table_name> WITH(type="database", propKey=”propVal”, … , Index='<attribute_name>')(<attribute_name> <attribute_type>, ...);
```
StoyanTc: @diana-macrometa - We cannot create inline indexes for `c8db` and you should also take out `Index='<attribute_name>'...`.

StoyanTc (commented):
@diana-macrometa - Why the `cli` is part of those changes?

docs/cep/table/index.md
diana-macrometa: What do you mean "inline indexes"?

Updated this line, btw.

diana-macrometa:
> @diana-macrometa - Why the `cli` is part of those changes?

Because I saw a bunch of backslashes that looked like a copy/paste error while I was searching for examples, so I fixed them.

docs/cep/table/index.md
StoyanTc: @diana-macrometa - The RDBMS allow the indexes to be created along with the table definition. Such indexes are named `inline`. However this is not valid for  `c8db`.

StoyanTc (approved):

PR #710, opened 2023-03-08
DOC-367: Publishing Data

diana-macrometa:
- Reworked the Data Publishing tutorial into an example.
- Moved some of the content from the original into the main sink topic.
- Updated Consuming Messages to be consistent with the new Publishing Data topic.
- Removed links to deleted file.


StoyanTc (approved):
@diana-macrometa - I did some reformatting changes. Let's avoid the horizontal scrollbars whenever is possible? This makes reading more comfortable and easy.

PR #717, opened 2023-03-13
Create Geospatial queries

diana-macrometa:
This PR creates a GeoSpatial Queries page and changes the order of a few other sidebar topics.

It also adds a file that contains a dataset.


ghost (approved):
👍

PR #758, opened 2023-03-21
DOC-702: Update optimizing-traversal-queries.md

diana-macrometa:
Major rework of this content. Please edit carefully.

@martin-macrometa - What is the query explainer? This topic reads like "query explainer" is a Macrometa thing, and I've never seen nor heard of it. The closest I can think of is the Execution Plan that appears in the Query Worker tab.


martin-macrometa:
EXPLAIN is a standard database operation (supported in all major SQL databases) which provides detailed information about how a query would be executed internally.
This is useful to estimate and improve performance - e.g. how many disk operations are necessary, what indices would be used

In our interface we renamed it to "Execution Plan" because somebody thought "Explain" didn't sound intuitive.

diana-macrometa:
> In our interface we renamed it to "Execution Plan" because somebody thought "Explain" didn't sound intuitive.

...and nobody updated the docs. Fabulous.

martin-macrometa (approved):

diana-macrometa:
@martin-macrometa, I will rewrite the docs for "Execution Plan" rather than "explainer" and then request a review again.

diana-macrometa:
@martin-macrometa, I updated the parts about the Execution Plan. Please re-review.

martin-macrometa (approved):

PR #764, opened 2023-03-22
DOC-701: Traversal examples

diana-macrometa:
Completely reworked the Traversal Examples topic. Lots of changed and new content, please edit carefully.


docs/graphs/graph-queries/traversal-queries/traversal-examples.md
```diff
@@ -3,179 +3,251 @@ sidebar_position: 40
...
-1. We start at the vertex **A**.
-2. We use a `min` depth of 1.
-3. We use a `max` depth of 2.
-4. We follow only in `OUTBOUND` direction of edges
+The traversal begins at vertex **A**:
 
 ![traversal_graph1](/img/traversal_graph1.png)
 
-Now it walks to one of the direct neighbors of **A**, say **B** (note: ordering is not guaranteed!):
+First it moves to a direct neighbor of **A**, for example, **B** (note: ordering is not guaranteed!):
```
martin-macrometa: "moves" doesn't sound right from the point of view of graph theory. "walk" is a technical term used for graph. "traverse", "search" and "visit" are also acceptable terms.

docs/graphs/graph-queries/traversal-queries/traversal-examples.md
```diff
@@ -3,179 +3,251 @@ sidebar_position: 40
...
+    "I",
+    "J",
+    "K"
+  ]
 ```
 
-```js
-    FOR v IN 2 OUTBOUND 'circles/A' GRAPH 'traversalGraph'
-      RETURN v._key
+The traversal starts at the first outer vertex, then returns to the branch and descends into the next tree. It repeats this process until all nodes have been visited. Both queries return the same results: the named graph query uses the graph name, while the collection sets query directly uses edge collections.
```
martin-macrometa: The query specified searches up to depth 3 from the start vertex. We should not be describing the order of traversal because this can vary depending on query options. I'm thinking of something along these lines:

The query visits all the neighbors of the specified vertex, their neighbors and so forth up to depth 3. Both queries return the same results: the named graph query uses the graph name, while the collection sets query directly uses edge collections.

docs/graphs/graph-queries/traversal-queries/traversal-examples.md
```diff
@@ -3,179 +3,251 @@ sidebar_position: 40
...
-- we know the `label` attribute of the edge connecting **A** to **G** is `right_foo`
+Suppose you want to exclude all vertices connected to **G**. There are several ways to do that.
 
-```js
-    FOR v, e, p IN 1..3 OUTBOUND 'circles/A' GRAPH 'traversalGraph'
-        FILTER p.vertices[1]._key != 'G'
-        RETURN v._key
+#### Filter by Vertex Key
+
+In this named graph query, the filter is applied to the vertex  `_key`.
```
martin-macrometa: In this named graph query, the filter is applied to the `_key` attribute of vertex 1 in the path. In particular, the query excludes the paths that go from the start vertex directly to vertex "G".

docs/graphs/graph-queries/traversal-queries/traversal-examples.md
```diff
@@ -3,179 +3,251 @@ sidebar_position: 40
...
+    RETURN v._key
 ```
 
-```js
-    FOR v, e, p IN 1..3 OUTBOUND 'circles/A' GRAPH 'traversalGraph'
-        FILTER p.edges[0].label != 'right_foo'
-        RETURN v._key
+#### Filter by Edge Label
+
+In this named graph query, the filter is applied to the edge label `right foo`.
```
martin-macrometa: the filter is applied to the label attribute of the start vertex of each edge

martin-macrometa (changes requested):

docs/graphs/graph-queries/traversal-queries/traversal-examples.md
diana-macrometa: I had no idea that "walk" was the technical term, I thought the original was just being folksy. I'll update it.

diana-macrometa: All edits applied.

docs/graphs/graph-queries/traversal-queries/traversal-examples.md
```diff
@@ -3,179 +3,251 @@ sidebar_position: 40
...
+This section provides an explanation of a simple graph traversal using an example. Each step of the traversal process is broken down as the query is executed.
+
+The query uses the following parameters:
+
+- Start at vertex **A**.
+- Use a `min` depth of 1.
+- Use a `max` depth of 2.
+- Traverse (move) only in `OUTBOUND` direction of edges.
+
+As discussed elsewhere, you can traverse edges and vertexes as a named graph or as a set of collections (anonymous graph, also called collection sets).
```
jamesm-macrometa: I would change "As discussed elsewhere" to a link to the section with the pertinent info, or just remove the line.

docs/graphs/graph-queries/traversal-queries/traversal-examples.md
```diff
@@ -3,179 +3,251 @@ sidebar_position: 40
...
+    "I",
+    "J",
+    "K"
+  ]
 ```
 
-```js
-    FOR v IN 2 OUTBOUND 'circles/A' GRAPH 'traversalGraph'
-      RETURN v._key
+The query visits all the neighbors of the specified vertex, their neighbors, and so on up to depth 3. Both queries return the same results: the named graph query uses the graph name, while the collection sets query directly uses edge collections.
```
jamesm-macrometa, line 132:
```suggestion
The query visits all the neighbors of the specified vertex, their neighbors, and so on up to depth 3. Both queries return the same results: The named graph query uses the graph name, while the collection sets query directly uses edge collections.
```

docs/graphs/graph-queries/traversal-queries/traversal-examples.md
```diff
@@ -3,179 +3,251 @@ sidebar_position: 40
...
+```json
+  [
+    "C",
+    "E",
+    "H",
+    "J"
+  ]
+```
+
+Understanding how to retrieve elements at a specific depth in a graph traversal is crucial for extracting meaningful insights from your data. By using the `min` and `max` depth parameters, you can effectively target the desired elements in your graph and simplify your query results.
```
jamesm-macrometa:
```suggestion
To extract meaningful insights from your data, you must understand how to retrieve elements at a specific depth in a graph traversal. By using the `min` and `max` depth parameters, you can effectively target the desired elements in your graph and simplify your query results.
```

docs/graphs/graph-queries/traversal-queries/traversal-examples.md
```diff
@@ -3,179 +3,251 @@ sidebar_position: 40
...
-    ]
-  ```
+## Refining Traversal with Filters
 
-As you can see, we can express this in two ways: with or without _max_ parameter in the expression.
+This section demonstrates how to apply filters to traversal queries for more precise results. The goal is to exclude specific branches or vertices from the graph. Two filtering approaches are illustrated below:
 
-## Filter examples
+- Filtering based on the vertex `_key` at depth 1, which is **G** in this case.
+- Filtering based on the label attribute of the edge connecting **A** to **G**, which is  `right_foo`.
```
jamesm-macrometa:
```suggestion
- Filtering based on the label attribute of the edge connecting **A** to **G**, which is `right_foo`.
```

docs/graphs/graph-queries/traversal-queries/traversal-examples.md
```diff
@@ -3,179 +3,251 @@ sidebar_position: 40
...
 
-Now let's start to add some filters. We want to cut of the branch on the right side of the graph, we may filter in two ways:
+### Filtering Branches Using Vertex Key or Edge Label
 
-- we know the vertex at depth 1 has `_key` == `G`
-- we know the `label` attribute of the edge connecting **A** to **G** is `right_foo`
+Suppose you want to exclude all vertices connected to **G**. There are several ways to do that.
```
jamesm-macrometa:
```suggestion
There are multiple ways to exclude all vertices connected to **G**.
```

jamesm-macrometa (approved):

docs/graphs/graph-queries/traversal-queries/traversal-examples.md
```diff
@@ -3,179 +3,251 @@ sidebar_position: 40
...
 
 ![traversal_graph1](/img/traversal_graph1.png)
 
-Now it walks to one of the direct neighbors of **A**, say **B** (note: ordering is not guaranteed!):
+First it walks to a direct neighbor of **A**, for example, **B** (note: ordering is not guaranteed!):
 
 ![traversal_graph2](/img/traversal_graph2.png)
 
-The query will remember the state (red circle) and will emit the first result **A** → **B** (black box). This will also prevent the traverser to be trapped in cycles. Now again it will visit one of the direct neighbors of **B**, say **E**:
+The query saves the state (red circle) and emits the first result **A** → **B** (black box). This also prevents the traverser from getting caught in cycles. Next, it proceeds to a direct neighbor of **B**, such as **E**:
```
martin-macrometa: This goes into specifics of the algorithm used. The actual algorithm can be configured via options. We should probably mention that the description assumes the default algorithm (depth-first search) is used.

I think a short description would be sufficient for people familiar with graphs:

The query performs a depth-limited depth-first search. The depth limit is specified by `max`.

diana-macrometa: How's this?

docs/graphs/graph-queries/traversal-queries/traversal-examples.md
```diff
@@ -3,179 +3,253 @@ sidebar_position: 40
...
+
+The query uses the following parameters:
+
+- Start at vertex **A**.
+- Use a `min` depth of 1.
+- Use a `max` depth of 2.
+- Traverse (move) only in `OUTBOUND` direction of edges.
 
-We use the following parameters for our query:
+You can traverse edges and vertexes as a named graph or as a set of collections (anonymous graph, also called collection sets).
```
koshy-thomas:
```suggestion
You can traverse edges and vertexes as a named graph or as a set of edge collections (anonymous graph, also called collection sets).
```

diana-macrometa: I didn't specify which collection, because we are talking about collections for both vertices and edges, which means in most cases, two types of collections: document and graph edge.

docs/graphs/graph-queries/traversal-queries/traversal-examples.md
```diff
@@ -3,179 +3,253 @@ sidebar_position: 40
...
 
 ![traversal_graph7](/img/graphs/traversal_graph7.png)
 
-After these steps there is no further result left. So all together this query has returned the following paths:
+Once these steps are completed, no further results remain. The query returns the following paths in total:
 
 1. **A** → **B**
-2. **A** → **B** → **E**
-3. **A** → **B** → **C**
+2. **A** → **B** → **C**
```
koshy-thomas: @diana-macrometa we will also might need to switch this in the explanation above too.. to first show and explain traversal from B->C and then from B-> E .. currently we explain B->E first and then B->C. It is a very minor comment just to align the doc.

diana-macrometa: Oh, THAT's why it was like that! I changed it to match my results. Good catch!

docs/graphs/graph-queries/traversal-queries/traversal-examples.md
koshy-thomas: No here we can only specify edge collections

diana-macrometa: Not really. A longer way of saying this would be:

You can traverse eges and vertexes as a named graph, which looks up your vertices and edges for you, or as a set of edge and document collections, where you specify the collections directly. The latter is called collection sets.

I do not think it would not be accurate to specify just edge collections here.

koshy-thomas: It is like edge collection are the basic thing required for collection sets you can also optionally specify vertex collections too. But the core concept is:
 Instead of GRAPH graphName you may specify a list of edge collections. Vertex collections are determined by the edges in the edge collections.

diana-macrometa: Oh, I get what you're saying now. The query only lists the edge collections. Got it!

koshy-thomas: I don't how to phrase this though but edgeCollections are mandatory for collection sets query to work.

diana-macrometa: I think that is covered in the syntax page.

koshy-thomas (approved):

PR #775, opened 2023-03-24
DOC-695: named vs anonymous graphs

diana-macrometa:
The concept of named graphs and anonymous graph is very important, so I took that content out of the Graph Queries index and put it in its own topic.

I also added additional explanation to the new topic while simplifying content in the old topic.


docs/graphs/graph-concepts/named-anonymous-graphs.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Anonymous Graphs
+
+_Anonymous graphs_, also known as _collection sets_, are another way to work with graphs in GDN. In this approach, you apply graph functions directly on a combination of document and edge collections, without explicitly defining a named graph. This method offers more flexibility and control over the collections involved in a graph and can be useful for ad-hoc graph queries or when working with multiple collections.
+
+With anonymous graphs, you have the freedom to mix and match collections as needed, making it easier to perform complex queries across various datasets. However, this approach requires a deeper understanding of the underlying collections and their relationships.
+
+## Which One to Use
+
+When deciding between named graphs and anonymous graphs, consider factors such as the complexity of your graph data, the frequency of ad-hoc queries, and the need for managing collections. Each approach has its advantages, and the choice depends on your specific requirements and preferences.
```
koshy-thomas: @diana-macrometa Here we are not exactly specifying the user when to choose what.. We should try to be more specific here when to choose what..

I found this in arangodb's docs we can take something from here.. 
`When to choose anonymous or named graphs?
As noted above, named graphs ensure graph integrity, both when inserting or removing edges or vertices. So you won’t encounter dangling edges, even if you use the same vertex collection in several named graphs. This involves more operations inside the database which come at a cost. Therefore anonymous graphs may be faster in many operations. So this question may be narrowed down to: ‘Can I afford the additional effort or do I need the warranty for integrity?’.`

docs/graphs/graph-concepts/named-anonymous-graphs.md
```diff
@@ -0,0 +1,28 @@
...
+
+Using named graphs simplifies the process of working with graph data because it allows you to focus on the graph structure itself, without worrying about the details of the underlying collections.
+
+## Anonymous Graphs
+
+_Anonymous graphs_, also known as _collection sets_, are another way to work with graphs in GDN. In this approach, you apply graph functions directly on a combination of document and edge collections, without explicitly defining a named graph. This method offers more flexibility and control over the collections involved in a graph and can be useful for ad-hoc graph queries or when working with multiple collections.
+
+With anonymous graphs, you have the freedom to mix and match collections as needed, making it easier to perform complex queries across various datasets. However, this approach requires a deeper understanding of the underlying collections and their relationships.
```
koshy-thomas: Can we also add something like this somewhere in this paragraph? It might be more clarifying, I guess :

`Anonymous graphs don’t have edge definitions describing which vertex collection is connected by which edge collection. The graph model has to be maintained in the client-side code. This gives you more freedom than the strict named graphs`

diana-macrometa: Great stuff! Added it in.

docs/graphs/graph-concepts/named-anonymous-graphs.md
```diff
@@ -0,0 +1,30 @@
...
+
+Using named graphs simplifies the process of working with graph data because it allows you to focus on the graph structure itself, without worrying about the details of the underlying collections.
+
+## Anonymous Graphs
+
+_Anonymous graphs_, also known as _collection sets_, are another way to work with graphs in GDN. In this approach, you apply graph functions directly on a combination of document and edge collections, without explicitly defining a named graph. This method offers more flexibility and control over the collections involved in a graph and can be useful for ad-hoc graph queries or when working with multiple collections.
+
+Anonymous graphs don’t have edge definitions describing which vertex collection is connected by which edge collection. The graph model has to be maintained in the client-side code. This gives you more freedom than the strict named graphs
```
koshy-thomas:
```suggestion
Anonymous graphs don’t have edge definitions describing which vertex collection is connected by which edge collection. The graph model has to be maintained in the client-side code. This gives you more freedom than the strict named graphs.
```

docs/graphs/graph-concepts/named-anonymous-graphs.md
koshy-thomas: just this one comment left for this PR

koshy-thomas (approved):

PR #776, opened 2023-03-24
DOC-693: Depth-first and breadth-first traversal

diana-macrometa:
This PR adds a new concept topic explaining depth-first and breadth-first traversal in the Macrometa context.


docs/graphs/graph-concepts/depth-first-breadth-first.md
```diff
@@ -0,0 +1,72 @@
...
+1. Visit the vertex at the front of the queue and mark it as visited.
+1. Add all unvisited adjacent vertices to the end of the queue.
+1. Remove the visited vertex from the queue.
+1. Repeat steps 2-4 until the queue is empty.
+
+BFT is particularly useful for tasks like finding the shortest path between two vertices in an unweighted graph or computing the minimum spanning tree of a connected graph.
+
+### BFT Example
+
+C8QL does not provide a built-in function for performing a breadth-first traversal explicitly. However, you can achieve a similar effect by using depth-first traversal and sorting the results by depth. Using the same `socialGraph` example, the query would look like this:
```
koshy-thomas: @diana-macrometa We can do BFT using bfs: true in OPTIONS

For example:

FOR vertex, edge, path IN 1..3 ANY 'circles/A' GRAPH 'traversalGraph'
OPTIONS {bfs: true}
  RETURN {
    user: vertex._key,
    connection: edge ? edge._key : null,
    depth: LENGTH(path.edges)
  }

docs/graphs/graph-concepts/depth-first-breadth-first.md
```diff
@@ -0,0 +1,72 @@
...
+1. Start at a source vertex.
+1. Visit the vertex and mark it as visited.
+1. For each adjacent unvisited vertex, perform a depth-first traversal from that vertex.
+1. Backtrack when there are no more unvisited adjacent vertices.
+
+DFT is particularly useful for tasks like searching for cycles in a graph or determining whether a path exists between two vertices.
+
+### DFT Example
+
+Suppose you have a graph named `socialGraph` with a vertex collection `users` and an edge collection `connections` representing the relationships between users. To perform a depth-first traversal starting from a user with the ID `users/12345` and explore up to a depth of 3, you can use the following C8QL query:
```
koshy-thomas: Also here it would be good if we can give a proper example.. maybe point to some preloaded example graphs we have like "traversalGraph" instead of "socialGraph", also it would be "circles" collection here instead of "users" and instead of "connections" it would be "circles" .. we can use "circles/A" as the starting vertex..
and the subsequent query will be:

`FOR vertex, edge, path IN 1..3 ANY 'circles/A' GRAPH 'traversalGraph'
  RETURN {
    user: vertex._key,
    connection: edge ? edge._key : null,
    depth: LENGTH(path.edges)
  }`

koshy-thomas: Basically just a suggestion to specify whatever sample data is to be used by the user to make the query running here , currently we don't specify how and what data to add in the respective collections

diana-macrometa: I honestly don't know how, lol. I just kept seeing those terms used, so I did some research, and here we are. I'm game for any improvements you can suggest.

docs/graphs/graph-concepts/depth-first-breadth-first.md
diana-macrometa: That's right! I'll take another stab at this.

docs/graphs/graph-concepts/depth-first-breadth-first.md
koshy-thomas: Okie Okie I will add that and modify the queries accordingly in the meanwhile you can rewrite the BFT thing as C8ql directly supports bft

diana-macrometa: Will do!

diana-macrometa:
@koshy-thomas, I have updated the content.

koshy-thomas (approved):

PR #777, opened 2023-03-24
DOC-694: Create pattern-matching-and-graphs.md

diana-macrometa:
This is brand new content explaining how to use pattern matching with GDN graphs.


docs/graphs/graph-concepts/pattern-matching-and-graphs.md
```diff
@@ -0,0 +1,46 @@
+---
+title: Pattern-matching and Graphs
+sidebar_position: 90
+---
+
+Pattern matching is a technique used to find specific structures or patterns within a graph. It enables expressing complex queries concisely and intuitively by defining a pattern to match in the graph data. Pattern matching is especially useful for identifying and extracting relevant subgraphs based on specific relationships between vertices and edges.
```
koshy-thomas, line 8: We are not exactly defining any pattern right...We are traversing the graph in some way trying to search for something..
 
 From what I found in arango:
 
 `If you traverse a graph to find paths fulfilling complex conditions then we call this kind of search, 'pattern matching'.`
 
 
 An example snippet I found from which we can probably extract and provide some more info:
 `We adventured pretty deep into the graph jungle
already. Exploring pattern matching in detail is
beyond the scope of this course, but let us take a
quick look at it nonetheless.
We can easily add filter conditions for the end
vertex and/or the edge which leads to it. Both are
emitted by the traversal as we know:
 FOR endVertex, edgeToVertex IN ...
With a variable traversal depth of 1..2 and the
default traversal options, there are 4 paths in the
following graph:
If we return the emitted end vertex, then the
result will contain the vertices A, B, C and C again.
We could also return the edges and would end up
with four edges in total. However, for the paths
S→A→C and S→B→C we may want to to choose
one over the other based on certain criteria. Full
paths can be optionally emitted as third variable:
 FOR vertex, edge, path IN ...
The path variable can then be used to apply filter
conditions on intermediate or all vertices and/or
edges on the path. 

This allows for queries like:
What are the best connections between the airports A
and B determined by the lowest total travel time?
It can be used to apply complex filter conditions in
traversals taking the entire path into account. 

In other words, it lets you discover specific patterns
– combinations of vertices and edges in graphs –
and is therefore called pattern matching.`

docs/graphs/graph-concepts/pattern-matching-and-graphs.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Traversal Operators
+
+- `->`: This operator represents a directed edge from the left-hand side vertex to the right-hand side vertex.
+- `<-`: This operator represents a directed edge from the right-hand side vertex to the left-hand side vertex.
+- `-`: This operator represents an undirected edge between the two vertices.
+
+## Example: Finding Posts by Followed Users
+
+Consider an example using a social network graph with two vertex collections, `users` and `posts`, and two edge collections, `follows` (representing users following other users) and `creates` (representing users creating posts). To find all posts created by users who are followed by a specific user, say `users/12345`:
```
koshy-thomas, line 20: If we can provide a sample data set , then it would be better from a user perspective

martin-macrometa (approved):

Luba89 (approved):

docs/graphs/graph-concepts/pattern-matching-and-graphs.md
diana-macrometa: This is supposed to be a concept topic, explaining the concept of pattern-matching and graphs. It's high-level, not instructional.

That said, if we want to make it more instructional, then I can move it to another place in the docs.

koshy-thomas (approved):

PR #780, opened 2023-03-27
DOC-328: Update Graph Collection index

diana-macrometa:
This PR adds information to the index of the Graph Edge collection section.

### DIFF
--- a/docs/collections/graph-edge/index.md
+++ b/docs/collections/graph-edge/index.md
@@ -3,7 +3,9 @@ sidebar_position: 1
 title: Graph Edge Collection
 ---
 
-An edge collection is a specialized type of collection that stores the relationships between vertices in a graph. These relationships are represented as directed or undirected edges, connecting two vertices. Edge collections have a unique structure that includes two special attributes: `_from` and `_to`.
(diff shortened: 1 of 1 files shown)


docs/collections/graph-edge/index.md
```diff
@@ -3,6 +3,13 @@ sidebar_position: 1
 title: Graph Edge Collection
 ---
 
-Graph Edge collections store your edges. For more information, refer to [Graphs](../../graphs/index.md).
+An edge collection is a specialized type of collection that stores the relationships between vertices in a graph. These relationships are represented as directed or undirected edges, connecting two vertices. Edge collections have a unique structure that includes two special attributes: `_from` and `_to`.
```
koshy-thomas: @diana-macrometa It doesn't necessarily store the relationships between vertices in a "graph". It can store relationship between any document of document collections. The corresponding documents will become the vertex and the corresponding collections will become the vertex collection but we can't say "graph" here I guess... I don't think we can specifically say graph here if we want to describe edge collection. 

Also it is always directed (_from -> _to) (never undirected).

diana-macrometa: We can't really avoid saying graph in this case, because the name of the collection in the UI is literally "Graph Edge."

koshy-thomas (approved):

PR #779, opened 2023-03-25
Graph Javascript SDK Example

ghost:
- Added Graph Javascript SDK example
- Explained example
- Explained recommendation query
- Added link to graph CRUD repository

PR #781, opened 2023-03-27
DOC-710: break up shortest path

diana-macrometa:
This PR breaks up the Shortest Path topic and adds use cases.


docs/graphs/graph-queries/shortest-path-queries/shortest-path-examples.md
```diff
@@ -0,0 +1,92 @@
...
+  RETURN [v._key, e._key]
+```
+
+This query specifies the edge collection to traverse (`edges`). The `OUTBOUND` keyword is used to traverse in the outbound direction from the start vertex `'circles/A'` to the target vertex `'circles/D'`. The `SHORTEST_PATH` keyword ensures that only the shortest path is returned.
+
+The `RETURN` statement returns an array containing the `_key` attributes of both the vertex (`v._key`) and the edge (`e._key`) of each step along the path.
+
+## Traverse Mixed Directions Example
+
+You can also compute shortest paths based on documents found in collections. You can use either the named graph syntax or the collection sets syntax, the results are identical. For information about running queries, refer to [Running Queries](../../../queries/running-queries).
```
koshy-thomas, line 38: @diana-macrometa here we have not specified an example plus the explanation  is also of the generic Shortest path example. 
We will need to add an example for this but the example will need multiple edge collection so the "traversalGraph" example which has been used in this whole page won't work for mixed directions example

diana-macrometa: Maybe the Cities Graph would work?

koshy-thomas: The problem with city Graph is that although there are three edgecollections but none of them have any common edges.. so adding any mixed direction traversal won't make any difference there. We will need to create some sample data but then the length will become big for this example.. anyways I will get you some working example for this tomorrow.. and then you can decide where to keep it and all

diana-macrometa: Check the example in Dino's SDK PR: https://github.com/Macrometacorp/docs/pull/779

koshy-thomas: Nope that won't work it uses a single edge collection

koshy-thomas: I will create a simple sample data for this today

docs/graphs/graph-queries/shortest-path-queries/shortest-path-examples.md
```diff
@@ -0,0 +1,92 @@
...
+
+### Named Graph
+
+```sql
+FOR a IN circles FILTER a._key == 'A'
```
Luba89: This query can be opt 
```
let a = (FOR a IN circles FILTER a._key == 'A' return a)[0]
let d = (FOR d IN circles FILTER d._key == 'D' return d )[0]
    FOR v, e IN 
      OUTBOUND SHORTEST_PATH a TO d 
      GRAPH 'traversalGraph' 
      RETURN [v._key, e._key]
```

docs/graphs/graph-queries/shortest-path-queries/shortest-path-examples.md
```diff
@@ -0,0 +1,92 @@
...
+```
+
+This query uses a named graph. It first filters the vertex with the key `'A'` from the `'circles'` vertex collection and assigns it to variable `'a'`. Similarly, it filters the vertex with the key `'D'` from the `'circles'` vertex collection and assigns it to variable `'d'`. It then finds the shortest path between vertex `'a'` and vertex `'d'` using the `'traversalGraph'` named graph.
+
+The `RETURN` statement returns an array containing the `_key` attributes of both the vertex (`v._key`) and the edge (`e._key`) of each step along the path.
+
+### Collection Sets
+
+```sql
+FOR a IN circles FILTER a._key == 'A'
```
Luba89: Same as above

docs/graphs/graph-queries/shortest-path-queries/shortest-path-examples.md
diana-macrometa: Updated both

docs/graphs/graph-queries/shortest-path-queries/shortest-path-examples.md
koshy-thomas: @diana-macrometa can we remove this section:
`## Traverse Mixed Directions Example

You can also compute shortest paths based on documents found in collections. You can use either the named graph syntax or the collection sets syntax, the results are identical. For information about running queries, refer to [Running Queries](../../../queries/running-queries).`

.. from shortest path examples for now? 
and later, we can add it you can assign a doc ticket to me saying to add a sample dataset and query because I was not able to find time for this today

diana-macrometa: I would prefer not to remove the section, because then it won't match the other query topics which have the Traverse Mixed Directions sections. Is it good enough for now? I mean, this content has been this way since forever. 

As long as everything is updated by the end of the week, it's all good.

koshy-thomas (approved):

docs/graphs/graph-queries/shortest-path-queries/shortest-path-examples.md
koshy-thomas: got it

docs/graphs/graph-queries/shortest-path-queries/shortest-path-syntax.md
```diff
@@ -0,0 +1,86 @@
...
+Object, _optional_.
+
+Used to modify the way the traversal runs. Only the following attributes have an effect, all others are ignored:
+
+- **weightAttribute** (string): A top-level edge attribute that should be used to read the edge weight. If the attribute is not existent or not numeric, then the _defaultWeight_ is used instead.
+- **defaultWeight** (number): This value is used as fallback if there is no _weightAttribute_ in the edge document, or if it's not a number. The default is 1.
+
+## Conditional shortest path
+
+The SHORTEST_PATH computation will only find an unconditioned shortest path. With this construct it is not possible to define a condition like: "Find the shortest path where all edges are of type _X_". If you want to do this, use a normal [Traversal](../traversal-queries/) instead with the option `{bfs: true}` in combination with `LIMIT 1`.
```
martin-macrometa, line 84: We should probably mention as a caveat that BFS can consume a lot of memory if the graph is very big.

diana-macrometa: More than we have in this concept topic? https://www.macrometa.com/docs/graphs/graph-concepts/depth-first-breadth-first

diana-macrometa: I actually updated the traversal queries topic just now, you're a reviewer.

PR #786, opened 2023-03-28
DOC-671: Create shortest-path.md

diana-macrometa:
Documentation for the SHORTEST_PATH keyword


docs/graphs/graph-queries/graph-keywords/shortest-path.md
```diff
@@ -0,0 +1,49 @@
...
+## Purpose
+
+The purpose of the `SHORTEST_PATH` keyword is to identify the most direct route between two vertices in a graph, considering the edge weights and directions. This allows users to efficiently find the minimum-cost paths within their graph data, enabling better decision-making, optimization, and analysis across various use cases and domains.
+
+## Syntax
+
+Here's the general syntax for using the `SHORTEST_PATH` keyword in C8QL:
+
+```sql
+FOR vertex, edge IN
```
koshy-thomas:
```suggestion
FOR vertex, edge IN OUTBOUND|INBOUND|ANY
```
Direction is mandatory

Also maybe for syntax we can link it to https://www.macrometa.com/docs/graphs/graph-queries/shortest-path-queries/shortest-path-syntax ??

docs/graphs/graph-queries/graph-keywords/shortest-path.md
```diff
@@ -0,0 +1,49 @@
...
+  GRAPH 'graph_name'
+  OPTIONS {weightAttribute: "attribute_name", defaultWeight: 1}
+  RETURN {vertex, edge}
+```
+
+## Query Parameters
+
+| Parameter         | Description         |
+|-------------------|---------------------------------------------|
+| start_vertex       | The starting vertex for the path search.                                |
```
koshy-thomas, line 32:
```suggestion
| IN OUTBOUND \| INBOUND \| ANY    |  defines in which direction edges are followed (outgoing, incoming, or both). |
| start_vertex       | The starting vertex for the path search.                                |
```
Should we specify the direction here also or not?

docs/graphs/graph-queries/graph-keywords/shortest-path.md
```diff
@@ -0,0 +1,49 @@
...
+| OPTIONS options    | (Optional) Additional options to customize the path search.             |
+| - weightAttribute  | (string) A top-level edge attribute that should be used to read the edge weight.   |
+| - defaultWeight    | (number) This value is used as fallback if there is no _weightAttribute_ in the edge document, or if it's not a number. The default is 1. |
+
+## Example
+
+Here's an example of finding the shortest path between two users in a social network graph:
+
+```sql
+FOR user, connection IN
```
koshy-thomas: After IN, a direction needs to be specified any one of OUTBOUND, INBOUND or ANY, 

We can also maybe simply link the example to this page ???:
https://www.macrometa.com/docs/graphs/graph-queries/shortest-path-queries/shortest-path-examples

docs/graphs/graph-queries/graph-keywords/shortest-path.md
```diff
@@ -0,0 +1,49 @@
+---
+title: SHORTEST_PATH
+---
+
+The `SHORTEST_PATH` keyword in [C8QL](../../../queries/c8ql/) is used to find the shortest path between two vertices in a graph. It computes the path with the fewest number of edges (hops) between a start vertex and a target vertex. This can be particularly useful for tasks such as finding the shortest route between two points in a transportation network, discovering the shortest chain of connections between two users in a social network, or analyzing the shortest dependency chain in a project management system.
```
koshy-thomas: Should we rephrase here or somewhere like to "single shortest path/route" basically, to be clear that it will only return one path even if if there are two paths with the same shortest length.

docs/graphs/graph-queries/graph-keywords/shortest-path.md
diana-macrometa: I thought that if you left it out, it would default to OUTBOUND.

I already linked to Shortest Path Queries, in the intro of this topic.

koshy-thomas (approved):

Luba89 (approved):

docs/graphs/graph-queries/graph-keywords/shortest-path.md
```diff
@@ -0,0 +1,52 @@
+---
+title: SHORTEST_PATH
+---
+
+The `SHORTEST_PATH` keyword in [C8QL](../../../queries/c8ql/) is used to find the shortest path between two vertices in a graph. It computes the path with the fewest number of edges (hops) between a start vertex and a target vertex. If it finds two paths with the same shortest length, then it returns only one.
```
martin-macrometa: By default the shortest path is calculated by the number of hops but a weight attribute can be specified in which case the shortest path would be the one with the smallest sum of weights over the edges in the path.

docs/graphs/graph-queries/graph-keywords/shortest-path.md
```diff
@@ -0,0 +1,52 @@
...
+```sql
+FOR user, connection IN OUTBOUND
+  SHORTEST_PATH 'users/12345' TO 'users/67890'
+  GRAPH 'socialGraph'
+  RETURN {user: user._key, connection: connection ? connection._key : null}
+```
+
+This query finds the shortest path between the user with the ID `users/12345` and the user with the ID `users/67890` in the `socialGraph`. The query returns an array of objects containing the user and connection information for each vertex and edge in the shortest path.
+
+It's important to note that `SHORTEST_PATH` returns the unweighted shortest path, meaning it only considers the number of hops (edges) between vertices. If you need to find the shortest path based on a specific attribute, such as distance or cost, you can use the `OPTIONS` parameter to specify a weight attribute and a custom weight function.
```
martin-macrometa: (related to my other comment)
This is confusing.

We should probably state that there are 2 ways to use `SHORTEST_PATH`. By default it counts the number of hops. Else, if `weightAttribute` is specified it sums weights.

martin-macrometa (changes requested):

docs/graphs/graph-queries/graph-keywords/shortest-path.md
diana-macrometa: Updated

docs/graphs/graph-queries/graph-keywords/shortest-path.md
diana-macrometa: I clarified that we're talking about just the example here.

martin-macrometa (approved):

PR #788, opened 2023-03-28
DOC-703: Document PRUNE keyword

diana-macrometa:
Adds documentation for the PRUNE keyword.

### DIFF
--- a/docs/graphs/graph-queries/graph-keywords/prune.md
+++ b/docs/graphs/graph-queries/graph-keywords/prune.md
@@ -10,7 +10,7 @@ For more information about using `PRUNE`, refer to [Pruning for Efficient Search
 
 ## Purpose
 
-The purpose of the `PRUNE` keyword is to control the traversal process by filtering out paths or branches during the graph traversal that do not meet specified conditions. This allows users to efficiently explore specific parts of a graph by pruning unwanted paths early in the traversal, reducing the overall complexity and improving query performance.
+The purpose of the `PRUNE` keyword is to control the traversal process by filtering out paths or branches during the graph traversal that do not meet specified conditions. This allows users to efficiently explore specific parts of a graph by pruning unwanted paths early in the traversal, reducing the overall complexity, improving query performance, and allowing you to manage resources such as memory and CPU usage more effectively.
 
 ## Syntax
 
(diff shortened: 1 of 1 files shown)


docs/graphs/graph-queries/graph-keywords/prune.md
```diff
@@ -0,0 +1,58 @@
...
+    PRUNE edge.relationshipType == 'blocked'
+    FILTER vertex.status == 'active'
+    RETURN vertex
+```
+
+In this example, the `socialNetwork` graph is traversed starting from the user with the ID `users/alice` and exploring the graph using an `OUTBOUND` direction. The traversal depth is set to a range between 1 and 4.
+
+The `PRUNE` keyword is used with the condition `edge.relationshipType == 'blocked'`. This means that if the traversal encounters an edge with a relationshipType attribute of `'blocked'`, then it does not explore that path further.
+
+Finally, the `FILTER` keyword is used to filter the results to include only vertices (users) with an `'active'` status. The query returns a list of active users reachable from `users/alice`, excluding paths with blocked relationships.
```
koshy-thomas, line 58: Just for clarification:
The paths up to this point (blocked relationship) will still be considered as result (they might still be post-filtered or ignored due to depth constraints). So the vertices, edge and path of the first edge it encounters with blocked relationshipType will still be considered in the result.

docs/graphs/graph-queries/graph-keywords/prune.md
```diff
@@ -0,0 +1,58 @@
...
+
+The PRUNE keyword in [C8QL](../../../queries/c8ql/) is used in conjunction with graph traversal queries to optimize performance by controlling the traversal process. It helps to reduce the number of paths explored during the traversal by stopping further exploration of a path if a specified condition is met.
+
+By using `PRUNE`, you can define criteria that, once met, will prevent the traversal from continuing along a particular path. This early termination of traversal can significantly improve query performance, especially in large graphs, by eliminating unnecessary exploration of nodes and edges that do not contribute to the desired result.
+
+For more information about using `PRUNE`, refer to [Pruning for Efficient Searches](../traversal-queries/traversal-pruning).
+
+## Purpose
+
+The purpose of the `PRUNE` keyword is to control the traversal process by filtering out paths or branches during the graph traversal that do not meet specified conditions. This allows users to efficiently explore specific parts of a graph by pruning unwanted paths early in the traversal, reducing the overall complexity and improving query performance.
```
Luba89: Maybe also add that helps manage resources such as memory and CPU usage more effectively.

Luba89 (approved):

docs/graphs/graph-queries/graph-keywords/prune.md
diana-macrometa: Updated the text. How does it look now?

docs/graphs/graph-queries/graph-keywords/prune.md
diana-macrometa: Updated. What do you think?

martin-macrometa (approved):

PR #810, opened 2023-03-30
uniqueVertices example

diana-macrometa:
This PR adds a new example topic with two use cases for the uniqueVertices traversal option.

### DIFF
--- a/docs/graphs/graph-examples/uniquevertices-examples.md
+++ b/docs/graphs/graph-examples/uniquevertices-examples.md
@@ -1,6 +1,6 @@
 ---
 sidebar_position: 100
-title: uniqueVertices Example
+title: uniqueVertices Examples
 ---
 
 This page provides examples of traversing graphs using the `uniqueVertices` option. For more information about this option, refer to [Traversal Query Syntax](../graph-queries/traversal-queries/traversal-syntax.md).
--- a/docs/graphs/graph-examples/uniquevertices-examples.md
+++ b/docs/graphs/graph-examples/uniquevertices-examples.md
@@ -18,7 +18,7 @@ The streaming platform has the following graph structure:
 
 ### C8QL Query
 
-The uniqueVertices option can be used in a [C8QL](../../queries/c8ql/) query to find unique viewers who have recently watched content from the 'Sci-Fi' genre:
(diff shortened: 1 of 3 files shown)


docs/graphs/graph-examples/uniquevertices-examples.md
```diff
@@ -0,0 +1,61 @@
...
+
+### C8QL Query
+
+The `uniqueVertices` option can be used in a [C8QL](../../queries/c8ql/) query to find unique viewers who have recently watched content from the 'Sci-Fi' genre:
+
+```sql
+FOR user, edge, path IN 1..1 OUTBOUND 'content/Sci-Fi' watched
+  OPTIONS {uniqueVertices: 'global'}
+  RETURN user
+```
```
Luba89, line 27: maybe something to add :
“global” – it is guaranteed that each vertex is visited at most once during the traversal, no matter how many paths lead from the start vertex to this one. If you start with a min depth > 1 a vertex that was found before min depth might not be returned at all (it still might be part of a path). Note: Using this configuration the result is not deterministic any more. If there are multiple paths from startVertex to vertex, one of those is picked. It is required to set bfs: true because with depth-first search the results would be unpredictable.

Luba89 (approved):

docs/graphs/graph-examples/uniquevertices-examples.md
Luba89: focus is on BFS

diana-macrometa: What do you mean by "focus is on BFS"?

Luba89: @diana-macrometa 
I didn't write it well, I was thinking that is good to add the following info:
Note: Using this configuration the result is not deterministic anymore. If there are multiple paths from startVertex to vertex, one of those is picked. It is required to set bfs: true because with depth-first search the results would be unpredictable.

diana-macrometa: This topic is an example, not the syntax documentation. The syntax documentation is linked in the page intro and is ALMOST exactly the same as what you have written here. I updated it with the missing bit, please take a look.

martin-macrometa (approved):

PR #839, opened 2023-04-17
Update tutorial.md

jgaonakm:
Remove a variable causing an error on the JS code sample

### DIFF
--- a/docs/queryworkers/tutorial.md
+++ b/docs/queryworkers/tutorial.md
@@ -135,7 +135,7 @@ const getCountValue = {
 };
 
 // Step 1: Open connection to GDN. You will be routed to the closest region.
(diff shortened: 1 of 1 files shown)


diana-macrometa:
Support guys, please review.

jesusgonzalez12:
@diana-macrometa, the thisApikey constant can be defined right after the globalUrl one on line 67 and then be referenced on line 92 with the client constant.
This is a way to show the user that they logged in with that API key.

If you think that's not necessary, deleting the `,  user: ${thisApikey}` is the way to go.

koshy-thomas (approved):

koshy-thomas:
Approved (there's no variable thisApiKey, and it's not needed) so we can go ahead with this!

koshy-thomas:
Thanks @jgaona-akamai !

diana-macrometa (approved):
Thanks for the fix!

PR #864, opened 2023-05-01
DOC-437: Document stream reader interface

diana-macrometa:
This PR documents the Reader interface for streams. It provides an explanation of what readers are, why you might want to use them, common tasks, and an example they can run in their console that leverages one of our sample stream workers.


docs/streams/stream-basics/consumers.md
```diff
@@ -26,10 +26,14 @@ Cumulative acknowledgement cannot be used with [shared subscription mode](subscr
 
 ## Readers
 
-GDN stream readers are like stream consumers, but they have two crucial differences:
+Stream readers are similar to stream consumers, but there are two crucial differences between them:
 
-- You can specify where on a stream readers begin processing messages; consumers always begin with the latest available unacknowledged message.
-- Readers don't retain data or acknowledge messages.
+- Readers allow you to specify the starting point for processing messages within a stream. In contrast, consumers always begin with the latest available unacknowledged message.
```
StoyanTc: The consumers could begin from either the `earliest` or the `latest` but read only **unacknowledged** messages. 
There are 2 main differences between readers and consumers:
1. The readers don't acknowledge messages.
2. The readers reader both acknowledged and unacknowledged messages.

StoyanTc (changes requested):

docs/streams/stream-tasks/manage-readers.md
```diff
@@ -0,0 +1,171 @@
+---
+sidebar_position: 110
+title: Manage Readers
+---
+
+Global Data Network (GDN) stream readers are similar to stream consumers, but there are two crucial differences between them:
+
+- Readers allow you to specify the starting point for processing messages within a stream. In contrast, consumers always begin with the latest available unacknowledged message.
```
koshy-thomas: Same point which Stoyan mentioned above. (earliest and latest unacknowledged message applies here \ too I guess)

koshy-thomas (approved):
Looks good to me!

StoyanTc (approved):

PR #860, opened 2023-04-27
DOC-528: improve unitconversion functions

diana-macrometa:
Continuing to improve function docs.

PR #868, opened 2023-05-02
Fix function mistakes

diana-macrometa:
Fixing mistakes in function documentation.


docs/cep/query-guide/functions/geospatial/crosses.md
```diff
@@ -24,25 +24,25 @@ Determines if the specified object or location crosses a geographic location spe
 ## Example 1
 
 ```sql
-CREATE STREAM StreamA (attribute_0 double, attribute_1 double, attribute_2 double, attribute_3 double);
-CREATE SINK STREAM OutputStream (attribute_0 double, attribute_1 double, attribute_2 double, attribute_3 double, prediction double, confidence double);
-
-@info(name = 'basicBayesianRegressionQuery')
-FROM StreamA#streamingml:bayesianRegression('model1', attribute_0, attribute_1, attribute_2, attribute_3)
-INSERT ALL EVENTS INTO OutputStream;
+SELECT geo:crosses('km-4354', -0.5, 0.5, '{"type":"Polygon","coordinates":[[[0, 0],[2, 0],[2, 1],[0, 1],[0, 0]]]}') AS crosses;
```
StoyanTc: @diana-macrometa - This is not a query technically but part of it. I would suggest to get rid of `SELECT` and `AS crosses`.
The below you could get rid of `This query uses`.

docs/cep/query-guide/functions/geospatial/crosses.md
```diff
@@ -24,25 +24,25 @@ Determines if the specified object or location crosses a geographic location spe
...
+This query uses the `geo:crosses()` function to determine if a line, defined by the coordinates `(-0.5, 0.5)` and `(km-4354)`, crosses the specified polygon. The polygon is represented as a GeoJSON string with coordinates `[[0, 0], [2, 0], [2, 1], [0, 1], [0, 0]]`.
 
 ## Example 2
 
 ```sql
-CREATE STREAM StreamB (attribute_0 double, attribute_1 double, attribute_2 double, attribute_3 double, attribute_4 double);
-CREATE STREAM DetailedOutputStream (attribute_0 double, attribute_1 double, attribute_2 double, attribute_3 double, attribute_4 double, prediction double, confidence double);
+CREATE STREAM InputStream (lineStart double, lineEnd double);
+CREATE STREAM OutputStream (crosses bool);
```
StoyanTc: @diana-macrometa - The same issue as yesterday with the `SINK` required for the `OutputStream`.

docs/cep/query-guide/functions/str/concat.md
```diff
@@ -20,10 +20,10 @@ This function returns a string value that is obtained as a result of concatenati
 
 ```sql
 @info(name = 'concatExample')
-SELECT concat('D533', '8JU^', 'XYZ') AS concatenatedString;
+SELECT str:concat('D533', '8JU^', 'XYZ') AS concatenatedString;
```
StoyanTc, line 23: @diana-macrometa - Why do you need this partial query for explaining the function? The query usage is anyway provided below. 
And this seems to be the same for all functions.

docs/cep/query-guide/functions/geospatial/crosses.md
diana-macrometa: My thought with this syntax is that we could give the user a line they could copy and paste into their own query.

docs/cep/query-guide/functions/str/concat.md
diana-macrometa: I thought the syntax would be helpful. Users could just copy and paste the line into their own query.

StoyanTc: @diana-macrometa - Yes, this makes sense but this is neither query since there is not `FROM` nor function. Eventually the customer would copy only `str:concat('D533', '8JU^', 'XYZ')` or add `FROM <table-name>`.

docs/cep/query-guide/functions/geospatial/crosses.md
```diff
@@ -33,7 +33,7 @@ This query uses the `geo:crosses()` function to determine if a line, defined by
 
 ```sql
 CREATE STREAM InputStream (lineStart double, lineEnd double);
-CREATE STREAM OutputStream (crosses bool);
+CREATE SINK STREAM OutputStream (crosses bool);
```
koshy-thomas: @diana-macrometa, the  example is incorrect,
An example for example:
```
CREATE STREAM dataInForGeoCrossesStream (id string, longitude double, latitude double, timestamp long);
CREATE SINK STREAM dataOutForGeoCrossesStream (id string, crosses  bool);
CREATE TRIGGER GeoCrossesTrigger WITH(interval = 5 sec);

INSERT INTO dataInForGeoCrossesStream
SELECT  "12" AS id, 6.876657 AS longitude, 79.897648 AS latitude,  eventTimestamp() AS timestamp
from GeoCrossesTrigger;

@info(name = 'geoCrossesQuery')
INSERT INTO dataOutForGeoCrossesStream
SELECT "id" AS id, crosses AS crosses
FROM dataInForGeoCrossesStream#geo:crosses("12", 0.5, 0.5,"{'type':'Polygon','coordinates':[[[0,0],[0,2],[1,2],[1,0],[0,0]]]}");
```
the first parameter "id" for geo:crosses .. needs to be a static string it seems can't be dynamic and this is a stream processor so it needs to be "FROM InputStream#geo:crosses"..
maybe @StoyanTc has a better example for can explain more..
The same is applicable to the Example 1 the standalone SELECT statement (i.e. also incorrect due to the above reason)

koshy-thomas: @StoyanTc can you also look at some comments mentioned for you here:
https://github.com/Macrometacorp/docs/pull/860
The unresolved comments

docs/cep/query-guide/functions/geospatial/crosses.md
koshy-thomas: @diana-macrometa still you cannot do SELECT geo:crosses .. It has to be "FROM InputStream#geo:crosses..." It's a stream processor not a function

docs/cep/query-guide/functions/str/concat.md
diana-macrometa: Are you saying I need to change all of these? Because over the last several PRs that you approved, this syntax appeared A LOT.

docs/cep/query-guide/functions/geospatial/crosses.md
diana-macrometa: Do I just need to change the SELECT to FROM?

koshy-thomas: Yes SELECT to FROM InputStream#geo:crosses....

koshy-thomas (approved):

PR #872, opened 2023-05-03
DOC-778: Stream worker ml topic

diana-macrometa:
This PR creates a topic to showcase our Streaming ML functions.


docs/cep/stream-worker-ml.md
```diff
@@ -0,0 +1,155 @@
...
+## Example 1: Bayesian Regression for Real-time Energy Consumption Prediction
+
+```sql
+CREATE STREAM TrainingStream (temperature double, time_of_day double, day_of_week double, recent_consumption double, energy_consumption double);
+CREATE STREAM PredictionStream (temperature double, time_of_day double, day_of_week double, recent_consumption double);
+
+CREATE SINK STREAM EnergyConsumptionPredictions (prediction double, confidence double, temperature double, time_of_day double, day_of_week double, recent_consumption double);
+
+@info(name = 'trainEnergyModel')
+INSERT INTO OutputStream1
```
StoyanTc: @diana-macrometa - `OutputStream1` is an internal stream. Such streams are used for intermediate data processing and finally have to transfer its data to either `sink` or `store`. 
Might be `OutputStream1` should be replaced by `PredictionStream` preliminary defined as `sink` . Otherwise `PredictionStream` is not used at all.
it's the same within the other samples below.

StoyanTc (changes requested):

docs/cep/stream-worker-ml.md
```diff
@@ -0,0 +1,155 @@
...
+SELECT *
+FROM TrainingStream#streamingml:updateBayesianRegression('energyModel', energy_consumption, temperature, time_of_day, day_of_week, recent_consumption);
+
+@info(name = 'predictEnergyConsumption')
+INSERT INTO EnergyConsumptionPredictions
+SELECT prediction, confidence, temperature, time_of_day, day_of_week, recent_consumption
+FROM PredictionStream#streamingml:bayesianRegression('energyModel', temperature, time_of_day, day_of_week, recent_consumption);
```
koshy-thomas: @diana-macrometa we should also update this page (https://www.macrometa.com/docs/cep/query-guide/functions/streaming-ml/bayesianregression) The example in the above link uses the old QL.. and plus the example won't work directly it needs updateBayesianRegression just like you used it here.. so maybe we can update the example in the bayesianregression page to something similar to what you have added here..

elof (approved):
I approve this as far as it's not breaking anything, and the content looks good. I did not test to see if the stream workers work though

diana-macrometa:
@elof, they passed the validator, but I like to wait for one or more of the SMEs to sign off before I merge. Especially because these are rather complex stream workers, and @koshy-thomas has occasionally been finding bugs in functions as he tests them.

docs/cep/query-guide/functions/streaming-ml/bayesianregression.md
```diff
@@ -30,16 +30,47 @@ streamingml:bayesianRegression(<STRING> model.name, <INT> prediction.samples, <D
...
+INSERT INTO OutputStream
+SELECT prediction, confidence, attribute_0, attribute_1, attribute_2, attribute_3
+FROM PredictionStream#streamingml:bayesianRegression('model1', attribute_0, attribute_1, attribute_2, attribute_3);
+```
+
+This query uses a Bayesian regression model named `model1` to predict the label of the feature vector represented by `attribute_0`, `attribute_1`, `attribute_2`, and `attribute_3`. The `StreamA` contains the input features, and the `PredictionStream` is created with the same features.
+
+A sink stream, `OutputStream`, is defined to store the predicted values, prediction confidence (std of predictive distribution), and the original features from the `PredictionStream`.
+
+The `predictAttributes` query processes events from the `PredictionStream`, using the Bayesian regression model 'model1' to make predictions based on the input features. The predicted values, along with their confidences and original features, are inserted into the `OutputStream` sink stream. As a result, the OutputStream stream is defined as follows: (prediction double, confidence double, attribute_0 double, attribute_1 double, attribute_2 double, attribute_3 double).
```
koshy-thomas: We cannot use Example 1 directly we need to add somewhere that before using any bayesianRegression model we need to initialise it first using updateBayesianRegression... and that's why example2 is good.

docs/cep/query-guide/functions/streaming-ml/bayesianregression.md
```diff
@@ -30,16 +30,47 @@ streamingml:bayesianRegression(<STRING> model.name, <INT> prediction.samples, <D
...
+INSERT INTO HousePricePredictions
+SELECT prediction, confidence, square_feet, num_bedrooms, num_bathrooms, age
+FROM HousePredictionStream#streamingml:bayesianRegression('housePriceModel', square_feet, num_bedrooms, num_bathrooms, age);
+```
+
+In this example, two input streams are created: `HouseDataStream` for model training and `HousePredictionStream` for making predictions. The `HouseDataStream` contains the square footage, number of bedrooms, number of bathrooms, age of the house, and the actual price. The `HousePredictionStream` contains the same features but without the actual price.
+
+A sink stream, `HousePricePredictions`, is defined to store the predicted house prices, prediction confidence, and the original features from the `HousePredictionStream`.
+
+The `trainHousePriceModel` query processes events from the `HouseDataStream`, updating the Bayesian linear regression model 'housePriceModel' using the input features and the actual house prices.
```
koshy-thomas:
```suggestion
The `trainHousePriceModel` query processes events from the `HouseDataStream`, and builds or updates the Bayesian linear regression model 'housePriceModel' using the input features and the actual house prices.
```

docs/cep/stream-worker-ml.md
```diff
@@ -0,0 +1,165 @@
...
+
+@info(name = 'trainEnergyModel')
+INSERT INTO ModelUpdateStatus
+SELECT *
+FROM TrainingStream#streamingml:updateBayesianRegression('energyModel', energy_consumption, temperature, time_of_day, day_of_week, recent_consumption);
+
+@info(name = 'predictEnergyConsumption')
+INSERT INTO EnergyConsumptionPredictions
+SELECT prediction, confidence, temperature, time_of_day, day_of_week, recent_consumption
+FROM PredictionStream#streamingml:bayesianRegression('energyModel', temperature, time_of_day, day_of_week, recent_consumption);
```
koshy-thomas: @StoyanTc After validating this SW if you click on Validate again it says the regression model already exists, please use a different name but shouldn't it "simply update the regression model" and not throw the error??
Second point, shouldn't it create the model only when the SW is published? and not when it's validated? also, just some more scenarios for regression models to think of is if the regression model is being deleted if the SW is unpublished.. or if it resumes after SW crashes and restarts..

StoyanTc: @koshy-thomas - We don't have experience with those extensions. 
Please file a ticket for investigating the protentional issue within.

diana-macrometa: I ran into that, too. I had to keep renaming the models as I tested the SWs to try and get them valid.

docs/cep/query-guide/functions/streaming-ml/bayesianregression.md
diana-macrometa: I removed the Bayesian examples and hid the documentation until the bug Koshy is going to file gets worked out.

diana-macrometa:
@StoyanTc, this is ready for your review.

docs/cep/stream-worker-ml.md
koshy-thomas: @StoyanTc / @diana-macrometa : Bug filed :
https://macrometa.atlassian.net/browse/CEP-562

koshy-thomas (approved):

diana-macrometa:
@StoyanTc, please review this PR

diana-macrometa:
Reviewers, please review.

diana-macrometa:
@StoyanTc, are you ready to unblock this?

StoyanTc (approved):

StoyanTc:
> @StoyanTc, are you ready to unblock this?

Yes, @diana-macrometa

PR #883, opened 2023-05-15
updated slack link

brandonLyonMM:
As requested per Slack https://macrometa-team.slack.com/archives/C036H7PMUTD/p1682970019789949

we are updating the URL for the community Slack server

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -130,7 +130,7 @@ const config = {
           },
           {
             className: 'navbar__item--external',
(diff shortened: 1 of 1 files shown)


james-macrometa:
@brandonLyonMM, there is a Slack Link in the footer as well that needs to be updated. Can you please get that one as well?

brandonLyonMM:
> @brandonLyonMM, there is a Slack Link in the footer as well that needs to be updated. Can you please get that one as well?

@jamesm-macrometa sorry I thought I did update the footer link. Now I don't know link I updated. That was the only reference I could find to Slack in the code. I'll keep looking.

PR #887, opened 2023-05-16
Update yarn.lock

diana-macrometa:
@brandonLyonMM, should this file be tracked in Git? It showed up as changed after I ran `yarn`.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -5140,10 +5140,15 @@ caniuse-api@^3.0.0:
     lodash.memoize "^4.1.2"
     lodash.uniq "^4.5.0"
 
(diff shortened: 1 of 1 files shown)


brandonLyonMM:
> @brandonLyonMM, should this file be tracked in Git? It showed up as changed after I ran `yarn`.

If the site still seems to run without issue, then it should be safe to commit these changes, since that was only a minor version bump.

The way that part of the package file is specified, it says "at least minor revision version 1.0", and when you ran yarn it grabbed the latest iteration of that version family, updated from 1.0.30000981 to 1.0.30001125.

PR #891, opened 2023-05-19
DOC-819: update quotas page

diana-macrometa:
This page adds data mesh integrations to the Quots and Limits page.


diana-macrometa:
Reviewers, please review.

docs/integrations/index.md
```diff
@@ -0,0 +1,39 @@
...
+
+Once created, you can reuse integrations as many times as necessary.
+
+## Data Source, Integration, Connection
+
+These terms are used frequently when discussing integrations.
+
+A _data source_ refers to an external database or entity that contains data, such as MongoDB, PostgresSQL, a different Macrometa collection, and so on.
+
+An _integration_ is a connection between Macrometa and a data source. You can create an integration with Macrometa.
```
grainier-macrometa, line 16: @diana-macrometa - `Integration` is more like a connection configuration to an external entity. 
@dustinlarimer / @elof - Correct me if I'm wrong here.

grainier-macrometa (approved):
> @diana-macrometa - Shall we move docs/integrations/integration-types/sql-server.md to docs/integrations/integration-types/mssql.md?

Do you mean change the file name?

docs/integrations/index.md
elof: @grainier-macrometa the way we were thinking about it before is that an integration is a group of ways to interact with other platforms and a connection is a type of integration. To add more confusion, I think we are going to move away from calling connectors integrations and just refer to them as connections. @dustinlarimer will be leading whatever changes we need here, but for now, I think the way @diana-macrometa described this will work

elof (approved):
:shipit:

diana-macrometa:
BTW folx, the Integrations intro was not supposed to be in this PR. It's still in progress, so you will see it again. Hopefully we'll have worked out the naming issue by then...

PR #892, opened 2023-05-19
Integration Ingestion modes

diana-macrometa:
This PR explains the different ingestion modes supported by integrations.

Reviewers, please feel free to expand on the content.

### DIFF
--- a/docs/integrations/integration-basics/ingestion-modes.md
+++ b/docs/integrations/integration-basics/ingestion-modes.md
@@ -13,10 +13,10 @@ LOG_BASED replication works by continuously checking these log files for new ent
 
 ## FULL_TABLE Ingestion
 
-This method works by scanning the entire table from the source database and copying all rows into the destination system. This happens each time the replication is run.
+This method works by scanning the entire table from the source database and copying all rows into the destination system. This happens when the replication is the first time.
 
-FULL_TABLE is the most comprehensive way of replicating data, as it ensures that the destination system contains a complete copy of the source table at the point the replication was run. However, it can be resource-intensive, particularly for large tables, and it doesn't reflect changes in real-time. Instead, the changes are only visible after the next replication run.
+FULL_TABLE is the most comprehensive way of replicating data, as it ensures that the destination system contains a complete copy of the source table at the point the replication was run. However, it can be resource-intensive, particularly for large tables, and it doesn't reflect changes in real-time. Instead, the changes are only visible if you reload the collection.
 
 ## Which to Choose
 
-Each of these methods has their pros and cons, and the choice between them often depends on the specific use case. For example, if real-time data is a requirement and the database supports it, LOG_BASED replication might be the better choice. However, if the database is relatively small and changes infrequently, FULL_TABLE replication could be simpler and just as effective.
+Each of these methods has their pros and cons, and the choice between them often depends on the specific use case. For example, if real-time data is a requirement and the database supports it, LOG_BASED replication might be the better choice. However, if you need to do a simple import of your full data, or manually update it infrequently, then FULL_TABLE replication could be simpler and just as effective.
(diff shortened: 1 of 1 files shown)


diana-macrometa:
Reviewers, please review.

docs/integrations/integration-basics/ingestion-modes.md
```diff
@@ -1,12 +1,22 @@
...
+## FULL_TABLE Ingestion
 
-  - Data Ingestion options - DFP - Can these be changed later? Saw these when setting up source at collection and when setting up target integration
-    - Batch Size
-    - Batch Flush Interval (Seconds)
-    - Batch Flush Minimum Time Gap (Seconds)
\ No newline at end of file
+This method works by scanning the entire table from the source database and copying all rows into the destination system. This happens each time the replication is run.
+
+FULL_TABLE is the most comprehensive way of replicating data, as it ensures that the destination system contains a complete copy of the source table at the point the replication was run. However, it can be resource-intensive, particularly for large tables, and it doesn't reflect changes in real-time. Instead, the changes are only visible after the next replication run.
```
grainier-macrometa: @diana-macrometa - `Instead, the changes are only visible after the next replication run`. For a given collection `FULL_TABLE` can be executed only once... So, I'm wondering whether this statement is correct.

docs/integrations/integration-basics/ingestion-modes.md
```diff
@@ -1,12 +1,22 @@
...
+
+## Which to Choose
+
+Each of these methods has their pros and cons, and the choice between them often depends on the specific use case. For example, if real-time data is a requirement and the database supports it, LOG_BASED replication might be the better choice. However, if the database is relatively small and changes infrequently, FULL_TABLE replication could be simpler and just as effective.
```
grainier-macrometa: @diana-macrometa - Here also, `FULL_TABLE` is a good choice if the user wants to completely migrate from some other DB or external service to MM regardless of the size or change frequency... And once moved, they can point to MM, instead of the external service. where `LOG_BASED` is more like a backup or a realtime synced replica of a external db/service...

grainier-macrometa (approved):
@diana-macrometa - Added a few comments. Other than that LGTM.

docs/integrations/integration-basics/ingestion-modes.md
diana-macrometa: ...I did not know that. So FULL_TABLE is a one-time thing? It won't capture ongoing changes?

I figured it would repeat on a timer or something.

docs/integrations/integration-basics/ingestion-modes.md
diana-macrometa: @elof, how do we want to handle this? We're getting into an interesting marketing/use case area, here. I'm pondering the messaging.

docs/integrations/integration-basics/ingestion-modes.md
grainier-macrometa: @diana-macrometa - Correct, `FULL_TABLE` is a one-time thing, and it won't capture ongoing changes. However, with `LOG_BASED` it will initially run a `FULL_TABLE` and then on top of that, use change logs to keep it in sync.

docs/integrations/integration-basics/ingestion-modes.md
elof: @diana-macrometa I think we should call FULL_TABLE a one time thing, and maybe say it's good for migrating data from a db to MM and LOG_BASED as a way to sync a third party DB to a MM collection

docs/integrations/integration-basics/ingestion-modes.md
diana-macrometa: Updated, please take a look.

elof (approved):
:shipit:

PR #898, opened 2023-05-26
DOC-782: Improve json functions

diana-macrometa:
This PR improves explanations for JSON functions in stream workers.


StoyanTc (approved):

docs/cep/query-guide/functions/json/getLong.md
```diff
@@ -38,8 +38,28 @@ Given a JSON object like `{'name' : 'John', 'age' : 23}`, the function `json:get
...
+@info(name = 'query3')
 json:getLong(json,'$.name')
 ```
 
 For a JSON object like `{'name' : 'John', 'age' : 23}`, the function `json:getLong(json,'$.name')` searches for a long value associated with the key `name`. In this case, it returns `null` as there is no matching long at `$.name`.
+
+## Example 4
+
+```sql
+CREATE STREAM PersonStream (json string);
```
diana-macrometa, line 50: @koshy-thomas and @Luba89, this one won't validate, cannot figure out why.

docs/cep/query-guide/functions/json/isExists.md
```diff
@@ -20,21 +20,54 @@ Function checks whether there is a JSON element present in the given path or not
...
+
+```sql
+json:isExists(json, '$.address.city')
+```
+
+For a JSON object like `{'name' : 'John', 'address' : {'city' : 'NY', 'country' : 'USA'}}`, the function `json:isExists(json,'$.address.city')` checks if an element exists at the key `address.city`. In this instance, it returns `true` because there is a matching element at `$.address.city`.
+
+## Example 5
+
+```sql
```
diana-macrometa, line 54: @koshy-thomas and @Luba89, also an issue with this one.

docs/cep/query-guide/functions/json/getLong.md
koshy-thomas: @diana-macrometa its bool not boolean, just change boolean to bool  in the line below this comment .. create sink MarriedPersonStream to: 
CREATE SINK STREAM MarriedPersonStream (name string, married bool);

docs/cep/query-guide/functions/json/isExists.md
koshy-thomas: @diana-macrometa this is also the same you need to change boolean to bool in the second line below this comment:
CREATE SINK STREAM AddressExistsStream (name string, addressExists bool);

koshy-thomas (changes requested):

koshy-thomas (approved):

PR #912, opened 2023-06-09
Update example of C8QL Queries for Doc Store

diana-macrometa:
This PR updates the topic that provides SDK code samples for using C8QL queries to work with docs.

**Reviewers**

The text returned for both of the SDK blocks was not very friendly or human-readable. I was able to simplify the Python responses, but not the JS. Maybe I simplified the Python too much? I'm not sure.

Please help make the responses match so that they have the same amount of detail and are human-readable.


docs/collections/documents/tutorials/c8ql-docs-queries-sdk.md
```diff
@@ -0,0 +1,247 @@
...
+          headers: self._headers,
+          body: body ? JSON.stringify(body) : undefined,
+          ...options
+        }).then((response) => self._handleResponse(response, resolve, reject));
+      });
+    }
+  }
+  
+  const apiKey = "XXXXX" // Use your API key here
+  const globalUrl = "api-play.paas.macrometa.io";
```
ghost: Can we just use one URL, for example, a federation URL, for this code sample? global = http = federation for this sample.

diana-macrometa: I do not understand what you mean. Can you make the edit as a suggestion and show me, or put it in a commit?

koshy-thomas: @diana-macrometa first of all you are not using the SDKs here at all.. the previous page that was deleted was only using APIs but the now this page says "SDK" but the examples are for direct API calls.. so I don't understand the requirement here.. you can remove term SDK id we want to show users c8ql queries with direct API calls... Please clarify this first..

diana-macrometa: Wait, what? I thought things like `sessions.headers.update` were from the SDK. If I am wrong, then I apologize and please let me know how to correct this.

diana-macrometa: The goal is to make this old piece of documentation relevant and useful for the current GDN.

diana-macrometa:
@koshy-thomas and @Luba89, please review. This PR is in violation of the new SLA agreement.

koshy-thomas:
@diana-macrometa I will make sure to find some time and provide the working sample codes and we can discuss it later today when you login and finally merge this PR

docs/collections/documents/tutorials/c8ql-docs-queries.md
```diff
@@ -0,0 +1,294 @@
...
+sidebar_position: 80
+title: C8QL Docs Queries Example
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+import Prerequisites from '../../../_partials/_prerequisites-sdk-api-key.md';
+import GetStarted from '../../../_partials/_get-started-steps.md';
+
+This page demonstrates how you can use C8QL and the Macrometa SDK to run CRUD operations on document store collection records. For more information about C8QL queries, refer to [C8QL](../../../queries/c8ql/) and [Queries](../../../queries/).
```
koshy-thomas: @diana-macrometa I have updated code samples to have the same output format but we need to change the above description where it says Macrometa SDK.. we are not using SDK in this example we are using APIs with Python and Javascript

koshy-thomas (changes requested):

docs/collections/documents/tutorials/c8ql-docs-queries.md
diana-macrometa: Updated. Anything else?

koshy-thomas: @diana-macrometa we also don't need Install SDKs Prerequisite:
image

diana-macrometa: Good catch! I changed them to the correct partials.

koshy-thomas (approved):

PR #913, opened 2023-06-12
Update stream-worker-tokenize.md

james-macrometa:
Removed ChatGPT response text


diana-macrometa (approved):
Thanks for catching this!

PR #914, opened 2023-06-12
Update query docs as api

diana-macrometa:
This PR updates the Query Docs as API to REST API Query Doc Store Example, mostly updating the login method to API key.


ghost (approved):
Great work @diana-macrometa 👍

diana-macrometa:
I would rather not merge the code until it is working. Any suggestions for how to fix it?

docs/collections/documents/tutorials/query-doc-store-api.md
```diff
@@ -1,40 +1,36 @@
...
+This page shows you how to query documents stored in a document store collection using the Macrometa API. For more information about using Macrometa APIs, refer to [APIs](../../../api-docs/index.md).
 
-## Prerequisites:
+## Prerequisites
 
-A Macrometa GDN tenant account and credentials.
+<Prerequisites />
 
-## Query as API (RestQL)
+## REST API Doc Query Example
```
koshy-thomas: @diana-macrometa First of all I guess there is a bit of confusion here.. This page here is for "Query Workers (which was earlier called RestQL)" using API (Rest api) calls but not "Query API itself" so maybe we need to change the title and definition of page??? 

I will fix the code samples provided here for "query using Query Workers (REST API Python and Js samples)" provided here and then we can have a call later to see what is the output format you want..

diana-macrometa: Sure! I am totally open to changing the title. This PR is about making this old documentation useful and relevant, so I am in favor of whatever needs to happen for that to be true.

diana-macrometa:
@koshy-thomas and @Luba89, please review. This PR is in violation of the new SLA agreement.

koshy-thomas:
@diana-macrometa I will make sure to find some time and provide the working sample codes today and we can discuss and change the title later today when you login and then we can finally merge this PR

docs/collections/documents/tutorials/query-doc-store-api.md
```diff
@@ -1,207 +1,258 @@
...
+title: REST API Query Doc Store Example
 ---
 
 import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
+import Prerequisites from '../../../_partials/_prerequisites-api-key.md';
+import Steps from '../../../_partials/_api-example-steps.md';
 
-Modern applications need to be highly responsive, always online, and able to access data instantly across the globe. At the same time, they need to be deployed on datacenters close to their users. Macrometa global data network (GDN) is a real-time materialized view engine that provides instant data to applications and APIs in a simple interface.
+This page shows you how to query documents stored in a document store collection using the Macrometa API. For more information about using Macrometa APIs, refer to [APIs](../../../api-docs/index.md).
```
koshy-thomas: @diana-macrometa I have updated the code samples to working ones with same output format.. But the thing is this whole section is about "QueryWorkers" using API, basically CRUD operations with QueryWorkers using API (python and javascript examples).. so we need to update the description/content/title of this page

koshy-thomas (changes requested):

docs/collections/documents/tutorials/query-doc-store-api.md
diana-macrometa: I updated the intro. Does the title also need changing?

Also, now that I understand this content better, I'm considering moving it to the Query Workers section. It seems like the "query worker" part of this example is more important than the fact that it is in a document collection. What do you think?

koshy-thomas: @diana-macrometa this whole example is doing CRUD operations on QueryWorkers and how you can use query workers using RestApi.. and not how you can use c8ql queries (yes in the end queryworkers are just queries but we want to focus on query worker part and yes we can move it to query workers section and rename everything)... I still don't feel the description is right.. But maybe you can have a fresh look from a Query Workers point of view

diana-macrometa: Took a stab at rewriting and updated the comments in the code to read "query workers" instead of "queries."

koshy-thomas: @diana-macrometa Looks good ..but what do you think abut the Heading/title of the page and the file name (.md)?

diana-macrometa: Do you have any suggestions?

elof: @diana-macrometa I checked this out. I think moving it and having the title be something like `REST API Query Worker create, read, update and delete` makes sense

diana-macrometa: Updated title and moved to Query Workers section.

koshy-thomas (approved):

PR #924, opened 2023-06-26
17.16 Stream region info

diana-macrometa:
This is not a 17.16 change, but it is something I just became aware of and it is newly important for connections, so I'm updating documentation to make sure it is reflected.

**Reviewers**: Please make sure that all of my changes are correct! I think they are based on my testing, but I would like them confirmed.


diana-macrometa:
Reviewers, please review.

koshy-thomas (approved):

PR #929, opened 2023-07-14
Doc 783 improve list function explanations

diana-macrometa:
A PR to add stream workers and improve list functions.


docs/cep/query-guide/functions/list/tokenize.md
```diff
@@ -27,8 +27,21 @@ list:tokenize(<OBJECT> list, <OBJECT> ...)
...
 ```sql
-@info(name = 'query1')
 list:tokenize(customList)
 ```
 
 The `list:tokenize(customList)` function takes a list named `customList`, and for each element in the list, it generates an event with the corresponding value. In this example, if `customList` contains elements (`gdn`, `IBM`, `XYZ`), the function will generate three events with value attributes `gdn`, `IBM`, and `XYZ` respectively.
+
+## Example 2
+
+```sql
```
diana-macrometa, line 37: @koshy-thomas and @Luba89, I got this error when I tried to validate the stream worker.

Error: Error on 'T-Bkuyr63ZGDeG2XqTUmVfGF:root:_system:test' @ Line: 8. Position: 24, near 'INSERT INTO OutputStream'. Different definition same as output 'stream definition OutputStream (index int, value object)' already exist as '@Sink( type = "stream", stream = "OutputStream", @Map( type = "json"))stream definition OutputStream (index int, value string)'

elof:
@diana-macrometa I reviewed the copy and checked that the pages are rendering correctly. If that's all you need for this review :shipit:

docs/cep/query-guide/functions/list/tokenize.md
StoyanTc: @diana-macrometa - Here the sink's value has to be `OBJECT`. Please give a try.

StoyanTc (approved):

docs/cep/query-guide/functions/list/tokenize.md
diana-macrometa: Yes, that did it. Thanks!

PR #932, opened 2023-07-17
Update Account Management Docs

diana-macrometa:
Updated Account Management docs due to an unexpected UI change that made it to production already.


elof (approved):
Looks good to me :shipit:

PR #939, opened 2023-07-27
17.17 - Semantic Search Index

diana-macrometa:
Add content to the index for the Semantic Search section in Search Views.

### DIFF
--- a/docs/search-views/fulltext-search/analyzers/index.md
+++ b/docs/search-views/fulltext-search/analyzers/index.md
@@ -7,8 +7,6 @@ _Analyzers_ enable you to break search inputs into sets of sub-values that searc
 
 You can use the [`TOKENS()` function](../queries/search-functions/tokens.md) to tokenize phrases and turn them into strings for C8QL search queries.
 
-An analyzer processes values based on its [features](features.md).
-
 ## Built-in Analyzers
 
 Macrometa provides a set of built-in analyzers.
--- a/docs/search-views/semantic-search/index.md
+++ b/docs/search-views/semantic-search/index.md
@@ -1,15 +1,11 @@
 ---
 sidebar_position: 1
-title: Semantic Search Views - Beta
+title: Semantic Search Views
 ---
 
 Semantic search is a technique that allows searches based on the intent and contextual meaning of search phrases, rather than just keyword matching. It understands the relationships between words and concepts to provide more relevant results.
 
 For example, a keyword search for "flights to Paris" would simply match documents containing those words. A semantic search would understand the intent is to find available flights to Paris and return pages for booking flights, rather than any page mentioning Paris and flights.
-
-:::note
-This is currently in beta.
-:::
 
 ## Benefits of Semantic Search
 
--- a/docs/search-views/semantic-search/index.md
+++ b/docs/search-views/semantic-search/index.md
@@ -5,6 +5,10 @@ title: Semantic Search Views
 
 Semantic search is a technique that allows searches based on the intent and contextual meaning of search phrases, rather than just keyword matching. It understands the relationships between words and concepts to provide more relevant results.
 
+:::note
+In 17.17, the Semantic Search Views work by finding similarities between search entities. Upcoming releases will add more nuanced and expanded functionality.
+:::
+
 For example, a keyword search for "flights to Paris" would simply match documents containing those words. A semantic search would understand the intent is to find available flights to Paris and return pages for booking flights, rather than any page mentioning Paris and flights.
 
 ## Benefits of Semantic Search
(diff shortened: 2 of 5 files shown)


docs/search-views/semantic-search/index.md
```diff
@@ -1,4 +1,31 @@
 ---
 sidebar_position: 1
-title: Semantic Search Views
+title: Semantic Search Views - Beta
```
amolth-macrometa: I think this is not beta.

docs/search-views/semantic-search/index.md
```diff
@@ -1,4 +1,31 @@
...
 ---
+
+Semantic search is a technique that allows searches based on the intent and contextual meaning of search phrases, rather than just keyword matching. It understands the relationships between words and concepts to provide more relevant results.
+
+For example, a keyword search for "flights to Paris" would simply match documents containing those words. A semantic search would understand the intent is to find available flights to Paris and return pages for booking flights, rather than any page mentioning Paris and flights.
+
+:::note
+This is currently in beta.
```
amolth-macrometa: Not sure about this.

docs/search-views/semantic-search/index.md
```diff
@@ -1,4 +1,31 @@
 ---
 sidebar_position: 1
-title: Semantic Search Views
+title: Semantic Search Views - Beta
 ---
+
+Semantic search is a technique that allows searches based on the intent and contextual meaning of search phrases, rather than just keyword matching. It understands the relationships between words and concepts to provide more relevant results.
```
amolth-macrometa, line 6: I think I gave this comment earlier (probably on a different PR). What you describe here is indeed correct description of semantic search, but I am concerned that we are not doing most of what is described here in 17.17. As a user I might feel let down by the actual score of 17.17 after reading this. Should we tone it down and focus more on similarity search which we are doing in 17.17?

amolth-macrometa (changes requested):

docs/search-views/semantic-search/index.md
diana-macrometa: Per the decision in Roadmap Ops a couple weeks ago, it is. Why do you say it is not? As I understand it, there is still a bunch of functionality to come before this is done.

docs/search-views/semantic-search/index.md
diana-macrometa: Yes, we absolutely should. We should only talk about what we can do right now. We can expand the description later, as we add functionality.

docs/search-views/semantic-search/index.md
amolth-macrometa: ok, I had heard otherwise from Durga hence the question.

docs/search-views/semantic-search/index.md
amolth-macrometa: ok, can you then modify this description to focus more on the similarity search?

docs/search-views/fulltext-search/analyzers/index.md
```diff
@@ -7,8 +7,6 @@ _Analyzers_ enable you to break search inputs into sets of sub-values that searc
 
 You can use the [`TOKENS()` function](../queries/search-functions/tokens.md) to tokenize phrases and turn them into strings for C8QL search queries.
 
-An analyzer processes values based on its [features](features.md).
```
maheshw2, line 10: Not sure why [Analyzer features](https://www.arangodb.com/docs/3.8/analyzers.html#analyzer-features) link is removed?

diana-macrometa: As I understand it based on the last time we chatted, the items listed in the Analyzer Features section were for custom analyzers, which we do not currently support. I hid that section until we do support them, so the link was broken due to the content being removed.

docs/search-views/semantic-search/index.md
diana-macrometa: I will get clarification.

docs/search-views/semantic-search/index.md
diana-macrometa: Literally nowhere in the UI do we talk about similarity search, so it's a little awkward to introduce a brand new concept here. Are we planning to add other searches besides similarity in the next release or two?

amolth-macrometa: You may not need to introduce a new term here, you can just say that in 17.17 the semantic search feature aims to find similarities between entities (or something like that) -  We can't avoid introducing the concept of similarity because that is essentially what we are doing here.

docs/search-views/fulltext-search/analyzers/index.md
maheshw2: This change LGTM. 
I'm not familiar with rest of the changes in PR.

docs/search-views/semantic-search/index.md
diana-macrometa: I like that! Added a note, please let me know if that works for you.

amolth-macrometa (approved):

PR #942, opened 2023-07-27
17.17 - Distance type concept

diana-macrometa:
This PR adds a concept topic explaining distance type.

### DIFF
--- a/docs/search-views/semantic-search/concepts/distance-type.md
+++ b/docs/search-views/semantic-search/concepts/distance-type.md
@@ -5,7 +5,7 @@ title: Distance Type
 
 This page provides basic information about values in the **Distance Type** field, which is used in defining semantic search views in Macrometa Global Data Network (GDN).
 
-In the context of semantic search, the "distance" between two entities is the extent of similarity between them. The greater the distance, the less similar the entities are.
+In the context of semantic search, the "distance" between two entities is the extent of similarity between them. The greater the distance, the less similar the entities are. A similarity search returns the entities which have the least distance from the queried entity.
 
 You have several distance type options. The different distance or similarity measures can have different impacts on search results, and the best one to use can depend on the specific characteristics of the data and the problem you're trying to solve.
 
--- a/docs/search-views/semantic-search/concepts/distance-type.md
+++ b/docs/search-views/semantic-search/concepts/distance-type.md
@@ -1,0 +1,20 @@
+---
+sidebar_position: 20
+title: Distance Type
+---
+
+This page provides basic information about distance type, which is used in defining semantic search views in Macrometa Global Data Network (GDN).
+
+You have several distance type options. The different distance or similarity measures can have different impacts on search results, and the best one to use can depend on the specific characteristics of the data and the problem you're trying to solve.
+
+## IP (Inner Product)
+
+The inner product (also known as the dot product) between two vectors is the sum of the products of their corresponding components. In the context of semantic search, inner product similarity is a measure of similarity that computes the dot product of the query vector and each document vector. High inner product values indicate high similarity. It's especially useful when dealing with vectors that represent things like user and item embeddings in recommendation systems.
+
+## L2 (Euclidean Distance)
+
+The L2 norm or Euclidean distance is probably the most commonly known metric. It represents the shortest straight line distance between two points in a space and is calculated as the square root of the sum of the squared differences between the corresponding elements of the two vectors. In high-dimensional spaces, this distance can become less intuitive due to the "curse of dimensionality", but it's often still a useful measure.
+
+## COSINE (Cosine Similarity)
+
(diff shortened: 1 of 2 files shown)


docs/search-views/semantic-search/concepts/distance-type.md
```diff
@@ -0,0 +1,20 @@
+---
+sidebar_position: 20
+title: Distance Type
+---
+
+This page provides basic information about values in the **Distance Type** field, which is used in defining semantic search views in Macrometa Global Data Network (GDN).
```
amolth-macrometa, line 6: At the start can you say that the "distance" between two entities is the extent of similarity between them. More the distance, less is the similarity. Similarity search essentially returns the entities which have the least distance from the queried entity. (Feel free to reword)

docs/search-views/semantic-search/concepts/distance-type.md
```diff
@@ -0,0 +1,20 @@
...
+
+## IP (Inner Product)
+
+The inner product (also known as the dot product) between two vectors is the sum of the products of their corresponding components. In the context of semantic search, inner product similarity is a measure of similarity that computes the dot product of the query vector and each document vector. High inner product values indicate high similarity. It's especially useful when dealing with vectors that represent things like user and item embeddings in recommendation systems.
+
+## L2 (Euclidean Distance)
+
+The L2 norm or Euclidean distance is probably the most commonly known metric. It represents the shortest straight line distance between two points in a space and is calculated as the square root of the sum of the squared differences between the corresponding elements of the two vectors. In high-dimensional spaces, this distance can become less intuitive due to the "curse of dimensionality", but it's often still a useful measure.
```
amolth-macrometa: maybe move this as the first distance type?

amolth-macrometa: Do we need to talk about this "curse of dimensionality" etc? I think these are concepts which have lot of literature available. I feel we should just explain these at high level.

amolth-macrometa (changes requested):

docs/search-views/semantic-search/concepts/distance-type.md
diana-macrometa: Great stuff, my only issue is that we don't use "similarity search" anywhere the user can see, so this sentence makes no sense: "Similarity search essentially returns the entities which have the least distance from the queried entity."

Could I replace "similarity" with "semantic" in that sentence and still have it be true?

docs/search-views/semantic-search/concepts/distance-type.md
diana-macrometa: Sure, but why?

diana-macrometa: I agree, I'm just so new to this sphere that I don't know how high or low is the right distance, so to speak. I appreciate the help fine-tuning it.

docs/search-views/semantic-search/concepts/distance-type.md
amolth-macrometa: no we could/should say that the semantic search functionality in 17.17 is limited to similarity search, and then talk about the similarity search. I agree we don't show the term much in the UI although the way to exercise this functionality is using the c8ql function named "similar".

docs/search-views/semantic-search/concepts/distance-type.md
amolth-macrometa: because it is described as the most commonly known metric in the text?

amolth-macrometa: yes I think this is too much information.

amolth-macrometa (approved):

docs/search-views/semantic-search/concepts/distance-type.md
diana-macrometa: Heh. Remember how in another PR you asked me about why I was calling this beta? It's because it doesn't seem like it's done yet! (Also, @romatastic  told me to.)

I'm thinking a note under the intro might be good. Are we planning to expand our semantic search beyond similarity search in the next release or two?

docs/search-views/semantic-search/concepts/distance-type.md
diana-macrometa: Works for me!

diana-macrometa: How does it look now?

docs/search-views/semantic-search/concepts/distance-type.md
amolth-macrometa: yes, I believe that we will keep expanding it. Semantic search is more of an umbrella than just a feature.

amolth-macrometa (approved):

docs/search-views/semantic-search/concepts/distance-type.md
diana-macrometa: Updated the text. What do you think?

amolth-macrometa (approved):

PR #959, opened 2023-08-11
17.17 Databricks Client Connector

diana-macrometa:
This PR adds Databricks Client Connector documentation. I have lots of questions, please refer to my comments.


docs/client-connectors/_category_.json
```diff
@@ -0,0 +1,4 @@
+{
+  "label": "Client Connectors - Beta",
```
diana-macrometa, line 2: Are client connectors a beta feature? I assumed they are, because connectors are.

docs/client-connectors/databricks-client/databricks-example.md
```diff
@@ -0,0 +1,76 @@
+---
+sidebar_position: 80
+title: Databricks Example
+---
+
+This example shows you how to read data from a Macrometa stream and write data to a Macrometa collection using the Macrometa Databricks Client Connector.
```
diana-macrometa, line 8: I would really like more context on this use case so that we can really show the reader why they might want to do this as well as how.

docs/client-connectors/databricks-client/index.md
```diff
@@ -0,0 +1,25 @@
+---
+sidebar_position: 1
+title: Databricks Client Connector
+---
+
+The Macrometa client connector for [Databricks](https://www.databricks.com/) enables you to seamlessly connect Macrometa [streams](../../streams/) and [collections](../../collections/) with Apache Spark within the Databricks environment. This comprehensive connector facilitates the ingestion, processing, and analysis of both streaming and batch data by leveraging Spark's advanced capabilities, allowing users to derive valuable insights and make data-driven decisions.
+
+The Macrometa Spark Connector offers two main components: Stream Data Connector and Collection Data Connector.
+
+## Stream Data Connector
```
diana-macrometa, line 10: Are the Stream Data Connector and Collection Data Connector really visible to the user? They seem like they might be kind of backend things.

docs/client-connectors/databricks-client/index.md
```diff
@@ -0,0 +1,25 @@
+---
+sidebar_position: 1
+title: Databricks Client Connector
+---
+
+The Macrometa client connector for [Databricks](https://www.databricks.com/) enables you to seamlessly connect Macrometa [streams](../../streams/) and [collections](../../collections/) with Apache Spark within the Databricks environment. This comprehensive connector facilitates the ingestion, processing, and analysis of both streaming and batch data by leveraging Spark's advanced capabilities, allowing users to derive valuable insights and make data-driven decisions.
```
diana-macrometa: Do we really want to talk about Apache Spark in our docs? @elof ?

docs/client-connectors/databricks-client/index.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Collection Data Connector
+
+The Collection Data Connector focuses on batch data processing to and from Macrometa collections.
+
+Source and target operations for collections can be executed using the format `com.macrometa.spark.collection.MacrometaTableProvider`.
+
+## Considerations
+
+- When mapping from a Macrometa array to a Spark array, Macrometa uses ArrayType. ArrayType which is a collection data type that extends the DataType class, which is a superclass of all types in PySpark. All elements of ArrayType should have the same type of elements.
```
diana-macrometa: I have no idea what this bullet means. Does it make sense?

docs/client-connectors/databricks-client/index.md
```diff
@@ -0,0 +1,25 @@
...
+- When inferring schema, the connector retrieves the first 50 documents from a collection and chooses the most common one between them.
```
diana-macrometa: Is this the same as the common schema comment in the Macrometa Collections connector?

docs/client-connectors/databricks-client/install-databricks-connector.md
```diff
@@ -0,0 +1,27 @@
...
+
+This page explains how to install the Databricks Client Connector.
+
+## Prerequisites
+
+- Databricks Runtime 11.3 LTS(with Apache Spark 3.3.0)
+- Scala 2.12 or later
+- A Macrometa account with write permissions for streams. For more information about permissions, refer to [Permissions](../../account-management/permissions/).
+
+## Install the Macrometa Databricks Client Connector
```
diana-macrometa, line 14: I reformatted this procedure quite a bit, please read carefully to make sure it still makes sense.

docs/client-connectors/databricks-client/read-from-collection.md
```diff
@@ -0,0 +1,53 @@
+---
+sidebar_position: 40
+title: Read from a Collection
+---
+
+The Macrometa Collections Databricks Connector allows you to integrate [Apache Spark](https://spark.apache.org/) with [Macrometa](https://www.macrometa.com/docs/collections/) collections, allows you to read data from Macrometa collections using Apache Spark.
+
+1. Set up your source options:
```
diana-macrometa, line 8: Where do I set up these source and target options? I get that there is a text file...somewhere...but I have no idea where I enter these parameters. Do we need to call that out, or will people using DataBricks just know automatically?

docs/client-connectors/_category_.json
grainier-macrometa: @diana-macrometa - I believe so. But let's check with @elof / @dustinlarimer / @rahulprabhune-macrometa first.

docs/client-connectors/databricks-client/databricks-example.md
grainier-macrometa: @diana-macrometa - Because Databricks functions as an external system, users can develop their own business logic using Spark within it. In this context, we serve as both a source (for taking inputs) and a target (for persisting outputs) from the perspective of Databricks. This example is divided into two sections: "Reading from a Macrometa Stream" and "Writing to a Macrometa Collection." These sections will demonstrate how to achieve only this using the client.

docs/client-connectors/databricks-client/index.md
grainier-macrometa: @diana-macrometa - In the context of Databricks, we function solely as a source and a target. The actual business logic will be written using Apache Spark within the Databricks environment. However, it might be prudent to seek @elof's perspective on this matter.

docs/client-connectors/databricks-client/index.md
grainier-macrometa: @diana-macrometa  - This will be easily noticeable to the end user. Using the `Stream Data Connector`, the user can continuously consume real-time events from a GDN stream, making it a persistent process. On the other hand, the `Collection Data Connector` enables querying of an existing GDN collection, which is a one-time action. Actually, these are not distinct connectors, but rather two distinct providers that users can alternate between.

docs/client-connectors/databricks-client/index.md
grainier-macrometa: @diana-macrometa - Correct, we use the same schema inference logic. Nevertheless, it's worth noting that the intended users of this connector might not be familiar with the `Macrometa Collections connector`. Hence, we might need to point that out here as well.

docs/client-connectors/databricks-client/index.md
grainier-macrometa: @diana-macrometa - Databricks notebooks support Python, Scala, and R. When using Python, Databricks utilizes PySpark internally. In this context, this statement is meaningful.

docs/client-connectors/databricks-client/install-databricks-connector.md
```diff
@@ -0,0 +1,27 @@
...
+
+Do not start until all prerequisites are fulfilled.
+
+1. Obtain the JAR file for the connector using one of the following methods:
+   1. **Using the official GitHub package (Recommended for production)**: Download the pre-built JAR file from the official GitHub package for this repository. For example: app-0.0.1.jar
```
grainier-macrometa: @diana-macrometa - Please link the Github packages to the first item. Link: https://github.com/Macrometacorp/macrometa-connector-databricks/packages/1908628

docs/client-connectors/databricks-client/install-databricks-connector.md
grainier-macrometa: cc: @praminda-macrometa @koshy-thomas

docs/client-connectors/databricks-client/read-from-collection.md
grainier-macrometa: @diana-macrometa - Typically, Databricks users will compose this code within the Databricks Notebook[1]. The `sourceOptions` represents a Map variable within that context. It's safe to assume that Databricks users are familiar with this setup.

[1] For more information, you can refer to: https://docs.databricks.com/en/notebooks/index.html

praminda-macrometa: @diana-macrometa This connector is used from a spark code users write on databricks platform. So each step below is decribing how someone should initialize the macrometa databricks connector within their spark source. There is no separate file or anything, users just have to use below lines of code inside their spark scripts to use our connector.
Simply think of below steps as explaining few lines of code users have to add into their source code.

grainier-macrometa (approved):
LGTM.

docs/client-connectors/databricks-client/install-databricks-connector.md
```diff
@@ -0,0 +1,27 @@
...
+   2. **Build from source**:
+      1. Clone this repository by running `git clone https://github.com/Macrometacorp/macrometa-connector-databricks.git`.
+      2. Build the JAR file using Gradle. Open a terminal in the root folder of the project and execute the command: `./gradlew clean shadowJar`. The generated JAR file, named 'macrometa-connector-databricks.jar', will be located in the `app/build/libs` directory.
+
+        This method provides the latest code, but it might not be officially released, so it's not recommended for production environments.
+
+2. Upload the JAR file to your Databricks workspace using the [Databricks CLI](https://docs.databricks.com/dev-tools/cli/index.html) or the Databricks UI.
+3. Attach the JAR file to your Databricks cluster by following the instructions in the Databricks documentation topic [Install a Library on a Cluster](https://docs.databricks.com/libraries/cluster-libraries.html#install-a-library-on-a-cluster).
```
praminda-macrometa: Let's use this link to make the link more focused. https://docs.databricks.com/en/libraries/workspace-libraries.html#install-workspace-libraries

docs/client-connectors/databricks-client/install-databricks-connector.md
praminda-macrometa: Added a comment. other than that, looks ok to me.

docs/client-connectors/databricks-client/databricks-example.md
diana-macrometa: Is this logic with Spark true for all client connectors? Or just Databricks?

grainier-macrometa: @diana-macrometa - Just Databricks

docs/client-connectors/_category_.json
diana-macrometa: @elof, @dustinlarimer - Please answer this question.

docs/client-connectors/databricks-client/index.md
diana-macrometa: @elof, please respond to the comments above.

diana-macrometa: Updated the text. What do you think?

docs/client-connectors/databricks-client/index.md
diana-macrometa: @elof, can you help me clarify? Or is it fine as-is from a customer perspective?

docs/client-connectors/databricks-client/index.md
diana-macrometa: Indeed, I was just going to reuse the write-up I have for the other collector, which has more detail.

diana-macrometa: Updated, please take a look.

diana-macrometa:
Please re-review, I added some significant content changes in index.md and want to make sure my updates in responses to your suggestions are satisfactory.

@elof, please take a look at questions aimed at you.

docs/client-connectors/_category_.json
elof: My opinion is that, if regular connectors are beta, than client connectors will also be considered beta. @diana-macrometa @grainier-macrometa @dustinlarimer

docs/client-connectors/databricks-client/index.md
elof: Yes, we can talk about Apache Spark in the context of Databricks, because Databricks is the commercial arm of Sparkk

docs/client-connectors/databricks-client/index.md
elof: @grainier-macrometa do we only support Python using PySpark, or will we also support other languages?

docs/client-connectors/databricks-client/index.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Collection Data Connector
+
+The Collection Data Connector focuses on batch data processing to and from Macrometa collections.
+
+Source and target operations for collections can be executed using the format `com.macrometa.spark.collection.MacrometaTableProvider`.
+
+## Mapping Considerations
+
+When mapping from a Macrometa array to a Spark array, Macrometa uses ArrayType. ArrayType which is a collection data type that extends the DataType class, which is a superclass of all types in PySpark. All elements of ArrayType should have the same type of elements.
```
koshy-thomas:
```suggestion
When mapping from a Macrometa array to a Spark array, Macrometa uses ArrayType. ArrayType which is a collection data type that extends the DataType class, which is a superclass of all types in Spark. All elements of ArrayType should have the same type of elements.
```

koshy-thomas: I missed to change this in original doc it seems.. we don't need to say `PySpark` we should just say `Spark`, we can say its a superclass of all types in `Spark`.(ArrayType with all elements of same data type is not specific to Pyspark but spark in general).

 In Macrometa, arrays can have elements with different data types but in Spark, ArrayType can have only elements that have the same datatype, so this point is basically to let the user know how the array datatype is converted between Macrometa and Spark.

docs/client-connectors/databricks-client/index.md
koshy-thomas: I missed to change this in the original doc it seems.. we don't need to say PySpark we should change `PySpark` to just `Spark`, we can say its a superclass of all types in Spark in general (ArrayType with all elements of same data type is not specific to Pyspark but Spark in general). 
In Macrometa, arrays can have elements with different data types but in Spark ArrayType can have only elements that have the same datatype, so this point is basically to let user know how the array datatype is converted between Macrometa and Spark.

praminda-macrometa (approved):

koshy-thomas (approved):

PR #966, opened 2023-08-18
DOC-877: Update semantic view

diana-macrometa:
This PR adds instructions for updating a semantic search view.


amolth-macrometa (changes requested):
You cannot update or rename a semantic search view. Properties specified at creation time are immutable.

diana-macrometa:
Good catch, @amolth-macrometa! I have deleted the topic.

amolth-macrometa (approved):

PR #971, opened 2023-08-23
Add  isCollectionStream Source option while reading from stream

koshy-thomas:


grainier-macrometa (approved):

diana-macrometa (approved):

diana-macrometa:
Nice catch, @koshy-thomas !

PR #973, opened 2023-08-28
add user defined schema step while reading

koshy-thomas:
Adding exmaple for user defined schema and correcting the info about the same in Readme.
The schema inference step is different for stream data connector.

### DIFF
--- a/docs/client-connectors/databricks-client/index.md
+++ b/docs/client-connectors/databricks-client/index.md
@@ -21,10 +21,15 @@ Source and target operations for collections can be executed using the format `c
 
 ## Mapping Considerations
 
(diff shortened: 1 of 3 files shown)


grainier-macrometa (approved):

diana-macrometa (approved):
Generally looks good, but I guess I need to update the databricks docs.

koshy-thomas:
@diana-macrometa by databricks docs if you mean this repo-
https://github.com/Macrometacorp/macrometa-connector-databricks/blob/master/GETTING_STARTED_WITH_COLLECTION_DATA_CONNECTOR.md

Then  its already synced

PR #977, opened 2023-09-06
Add 17 16 api links

diana-macrometa:
This PR adds API links to Connections docs.

**Reviewers**

Please pay special attention to the part about adding transformations. I'm not sure the steps are in the right order, and I could not figure out which endpoint actually adds the transformation to the connection.

### DIFF
--- a/docs/connections/manage-etl-workflows.md
+++ b/docs/connections/manage-etl-workflows.md
@@ -47,6 +47,7 @@ To add a transformation:
 2. [Generate a Transformation Query Sample](https://www.macrometa.com/docs/api#/operations/getSampleTransformation)
 3. [Validate the Transformation](https://www.macrometa.com/docs/api#/operations/validateTransformation)
 4. [Generate Transformation Preview](https://www.macrometa.com/docs/api#/operations/previewTransformation)
+5. Add the transformation query in the [Create a Source or Target Workflow](https://www.macrometa.com/docs/api#/operations/createWorkflow) request.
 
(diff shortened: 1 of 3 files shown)


docs/connections/manage-etl-workflows.md
```diff
@@ -43,21 +71,48 @@ You can only add target workflows to a Macrometa document collection that contai
...
 
+</TabItem>
+<TabItem value="api" label="REST API">
+
+Use our interactive API Reference with code generation in 18 programming languages to [Create a Source or Target Workflow](https://www.macrometa.com/docs/api#/operations/createWorkflow).
+
+To add a transformation:
+
+1. [Retrieve Sample Data from the Data Source](https://www.macrometa.com/docs/api#/operations/retrieveSample)
+2. [Retrieve the Schema from the Data Source](https://www.macrometa.com/docs/api#/operations/retrieveSchema)
```
praminda-macrometa: @diana-macrometa Step 1. response contain both sample data and the schema. So Step 2. Is not a mandatory step in tranformation creation. Everything else is correct in this flow

diana-macrometa: But how do we add the transformation to the workflow?

praminda-macrometa: @diana-macrometa we have to provide the transformation query created in here to `createWorkflow` request.

praminda-macrometa (approved):

PR #979, opened 2023-09-08
17.17: Add KV group documentation

diana-macrometa:
This PR adds documentation for KV Group ID changes.


diana-macrometa:
@SpringNgMacrometa, this PR is one day away from breaching SLA.

docs/collections/keyvalue/index.md
```diff
@@ -8,6 +8,33 @@ import TabItem from '@theme/TabItem';
...
+
+When you create a key-value store collection, you can select the **Group** option to enable the **GROUP ID** field. This allows you to organize key-value pairs in groups. This option has several effects:
+
+- The Group ID field is automatically indexed, which makes querying based on the ID faster.
+- You can perform queries and truncate based on the group ID value.
+
+Here are some example queries:
+
+```sql
+FOR d IN <groupID> RETURN d;
```
akash-macrometa: Both the queries are not working.
Second one is a C8QL query. First one does not seem a valid query.

docs/collections/keyvalue/key-value-store-data.md
```diff
@@ -29,6 +29,10 @@ The key that corresponds to the document. Every key has a unique value.
 
 Displays the value corresponding to the key.
 
+## Group ID
```
akash-macrometa, line 32: If no Group ID was given while creating or modifying the document, we will see "N/A", Not sure if it should be mentioned.

static/img/collections/truncate-group-id.png
akash-macrometa, line 1: If we pass the group ID, all the documents will not be deleted, only documents belonging to that particular group will be deleted. Sorry, I missed it earlier. cc @VikasPatel

akash-macrometa (changes requested):

docs/collections/keyvalue/index.md
akash-macrometa: A SQL query should look like `select _key, value from <collection> where groupID = <groupID"`

diana-macrometa: I got the queries from the QA handoff video, FYI.

diana-macrometa: For C8QL queries, I generally tag them as SQL for markdown rendering, because it doesn't know what to do with C8QL. Can you correct them for C8QL?

docs/collections/keyvalue/key-value-store-data.md
diana-macrometa: Good info, doesn't hurt to add it.

static/img/collections/truncate-group-id.png
diana-macrometa: Um, does this mean the image is wrong? Do I need to make an update here?

docs/collections/keyvalue/index.md
akash-macrometa: First one is incomplete, no mention of collection. You must remove that one. The second one is correct C8QL. However, it should not end with  `;`. 
You can add one more ```FOR d IN <collection> FILTER d.groupID == "<groupID>" RETURN {_key: d._key, value: d.value}```

static/img/collections/truncate-group-id.png
akash-macrometa: Yeah, I think UI team will have to make changes. 
The behaviour is, If you do not pass any groupID, it will delete all the documents in the collection. But if the groupID is given, only documents with the mentioned groupID will be deleted.

docs/collections/keyvalue/index.md
diana-macrometa: Updated

akash-macrometa (approved):

PR #1013, opened 2023-10-19
Create vwrs-metrics.md

diana-macrometa:
Creating a page for VWRS metrics.


docs/photoniq/vwrs/vwrs-metrics.md
```diff
@@ -0,0 +1,73 @@
...
+You can retrieve metrics using the following endpoints:
+
+- Get Metrics for a Domain - Retrieve the metrics for a specific domain. Set start and end times to get metrics for a specific time period.
+- Get All Metrics - Retrieve metrics for all domains. Set start and end times to get metrics for a specific time period.
+
+## Available Metrics
+
+The metrics response body includes the following properties:
+
+- **avg_waiting_time**: (integer) The average time users spend in the virtual waiting room before being granted access to the origin service (in milliseconds).
```
joubert-macrometa: I would remove the (type) after each description.  This is a very explicit thing that probably is not needed here.

docs/photoniq/vwrs/vwrs-metrics.md
```diff
@@ -0,0 +1,73 @@
...
+  {
+    "_key": "www.example1.com:2023-08-21_07:50:00",
+    "avg_waiting_time": 0,
+    "domain_key": "www.example1.com",
+    "duplicate_request_rate": 100,
+    "peak_queue_length": 100,
+    "queue_abandonment_rate": 0,
+    "request_success_rate": 100,
+     "usage_requests": 6000,
+    "timestamp": 1692604200000
```
joubert-macrometa: Strange.  The timestamp is in milliseconds.  I assume you copied this from somewhere.  We should fix this there too.
Remove the last three zeros from the timestamp to get it in seconds (as defined above).
You can verify the timestamp here: https://www.epochconverter.com/

All three samples have the timestamp in milliseconds.  Remove the last three zeros to get it to seconds for all of these.
@rahul-naik1 does our swagger file contain these examples?  If so, we should probably fix the timestamp.

joubert-macrometa (changes requested):

docs/photoniq/vwrs/vwrs-metrics.md
diana-macrometa: I copied this example from the Box file.

docs/photoniq/vwrs/vwrs-metrics.md
```diff
@@ -0,0 +1,73 @@
+---
+sidebar_position: 100
+title: Get VWRs Metrics
+---
+
+Macrometa PhotonIQ Virtual Waiting Rooms (VWRs) has built-in metrics that allow you to retrieve waiting room telemetry per waiting room, per minute. Data is stored for the past 90 days.
```
joubert-macrometa, line 6: This got me thinking:  VWR vs VWRS.
What is the S?  "Get the VWRs metrics" ==> "Get the Virtual Waiting Room metrics"  What happened to the "s"

I think originally the S was Service ==> Virtual Waiting Room Service (is that right @rahul-naik1)

Should this then start with "Macrometa PhotonIQ Virtual Waiting Room Service..."
and then what about  "Rooms" 

@diana-macrometa you know, me overthinking probably, but thought I would just say it out loud.  You do with it what you think is best.

diana-macrometa: The "s" just makes it plural.

diana-macrometa: @nicolewilliams88 and @macrochrissie , any comment?

joubert-macrometa (approved):

docs/photoniq/vwrs/vwrs-metrics.md
rahul-naik1: Let me check what we returning

rahul-naik1: @joubert-macrometa
We are storing values in milliseconds

docs/photoniq/vwrs/vwrs-metrics.md
rahul-naik1: @joubert-macrometa Yes, it is for Service

PR #1032, opened 2023-11-07
18.00 Instances Metadata Update

diana-macrometa:
This PR adds content for [CEP-617](https://macrometa.atlassian.net/browse/CEP-617) to the documentation.

@StoyanTc, please take a close look at the new example! I didn't see a code example in the ticket, so I am totally guessing and it is probably wrong. Please also make sure that the content is correct. The ticket was sparse, so I am not confident in the text I wrote.

[CEP-617]: https://macrometa.atlassian.net/browse/CEP-617?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/cep/metadata.md
+++ b/docs/cep/metadata.md
@@ -56,6 +56,16 @@ For example, this annotation would run three instances of the stream worker:
 @App:instances('3')
 ```
 
+In some cases, you might need to distribute instances of the same stream worker across multiple nodes to have more scalability.
+
+`@App:instances()` metadata has a second optional parameter: `distributed  = true/false`. If `true`, then the instances are run on multiple nodes. If `false`, then they all run on the same node. Default is `true`.
+
+For example, this annotation would run three instances of the stream worker on the same node:
+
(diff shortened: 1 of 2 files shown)


StoyanTc (commented):
No changes are required into the documentation but only into the release notes.
The parameter `distributed` wasn't implemented as redundant.

diana-macrometa:
Then what changed? Do we just always run stream workers on multiple nodes automatically?

StoyanTc:
> Then what changed? Do we just always run stream workers on multiple nodes automatically?

The internal rebalancing of the stream workers among the nodes was changed, i.e. all instances of given stream worker were run only on one node before and after the fix are subject of evenly distribution on all available node.

diana-macrometa:
Updated.

StoyanTc (approved):

PR #1041, opened 2023-11-20
18.00 api endpoint button

diana-macrometa:
This PR adds documentation for the new API Endpoint button in stream workers. ([GUI-1645](https://macrometa.atlassian.net/browse/GUI-1645))

**Reviewers**

I'm not clear on how or why this API endpoint is useful to the user. How will they use it? What does the user need to know?

[GUI-1645]: https://macrometa.atlassian.net/browse/GUI-1645?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/cep/stream-worker-tasks/create-api-endpoint.md
```diff
@@ -0,0 +1,23 @@
...
+1. Write and save stream worker as described in [Create a Stream Worker](create-stream-worker.md).
+1. Click **API Endpoint**.
+
+Macrometa displays the cURL (Client URL) command that you can use in the terminal or command line to make an API call and run the query.
+
+![API Endpoint](/img/stream-workers/api-endpoint.png)
+
+## Stream Worker Endpoints
+
+Why would a user want to do this? Can they send info to a stream worker? Receive it from? What is the point of this cURL commend?
```
grainier-macrometa: cc: @StoyanTc @dlazarenko-macrometa

StoyanTc: @VikasPatel - I wasn't aware for this new button `API Endpoint` till now. This seems to be on place of `Edge workers` but how the user could use it?
image

diana-macrometa:
Here is what @dwivedianurag said in the ticket: In query worker editor we have a button API endpoint on clicking which we display the curl command. Similarly add the button in stream worker editor as well. On this button upon clicking please display the curl command for publish  message stream worker Rest api

docs/cep/stream-worker-tasks/create-api-endpoint.md
diana-macrometa: @dwivedianurag, can you answer this?

StoyanTc: Some additional info, this button works only at valid stream worker's definition and generates a curl statement for publishing.
I really don't see the real usage of it.

diana-macrometa: @rahulprabhune-macrometa, if it is not useful, then how hard would it be to remove this?

cc @romatastic, @durgagokina

StoyanTc: It's not consistent in my opinion, the user could ask if this is for `publishing` how could I get for `unpublishing` using the same or another button.

docs/cep/stream-worker-tasks/create-api-endpoint.md
```diff
@@ -20,4 +20,4 @@ Macrometa displays the cURL (Client URL) command that you can use in the termina
 
 ## Stream Worker Endpoints
 
-Why would a user want to do this? Can they send info to a stream worker? Receive it from? What is the point of this cURL commend?
+The generated cURL command provides an easy way to get a JavaScript Web Token (JWT) and the URL to the API endpoint. This can be especially useful to call the API for testing purposes.
```
StoyanTc, line 23: This is only one of the many API Stream Workers endpoints - that for `publishing`. Shouldn't we specify it?

StoyanTc (approved):

PR #1049, opened 2023-11-28
18 00 transactions index

diana-macrometa:
This PR is the beginning of Transactions documentation. More will be added in future PRs, this is just an introduction page.

### DIFF
--- a/docs/transactions/index.md
+++ b/docs/transactions/index.md
@@ -2,3 +2,44 @@
 sidebar_position: 1
 title: Transactions
 ---
+
+Macrometa provides support for user-definable transactions within a region. _Transactions_ allow you to perform multi-document transactions with individual begin, commit, abort commands. They work similar to the BEGIN, COMMIT, and ROLLBACK operations in relational database systems.
+
+## How They Work
+
+Simply put, transactions are a way to aggregate, or group, multiple API calls. The process is as follows:
+
+### 1. Create a Transaction
+
+When you create the transaction, you specify which collections the transaction will write to.
+
+Once you create a transaction, you will be given a transaction reference number.
+
+### 2. Make API Calls
+
+Make API calls that include the transaction reference number to include them in the transaction. All actions will run, but they are not final until you commit the transaction.
+
+### 3. Commit or Abort Transaction
+
+- To finalize all transaction API calls, commit the transaction.
+- To roll back the actions, perhaps because one of them failed, abort the transaction.
+
+## ACID Properties
+
+Transactions in Macrometa within a region are atomic, consistent, isolated, and durable (ACID).
+
+These ACID properties provide the following guarantees:
+
+- The _atomicity_ property makes transactions either complete in their entirety or have no effect at all.
+- The _consistency_ property ensures that no constraints or other invariants will be violated during or after any transaction.
+- The _isolation_ property hides the modifications of a transaction from other transactions until the transaction commits.
+- The _durability_ property ensures that operations from transactions that have committed will be made persistent.
+
+## Supported Operations
+
+Supported operations include:
+
+- Read and write documents
+- Get the number of documents of collections
+- Truncate collections
+- Run queries or query workers
--- a/docs/transactions/index.md
+++ b/docs/transactions/index.md
@@ -32,7 +32,7 @@ Supported operations include:
 - Get the number of documents in a collection.
 - Get the number of key-value pairs in a collection.
 - Truncate collections.
-- Run queries or query workers.
+- Run queries.
 
 ## ACID Properties
 
--- a/docs/transactions/index.md
+++ b/docs/transactions/index.md
@@ -13,11 +13,11 @@ Simply put, transactions are a way to aggregate, or group, multiple API calls. T
 
 When you create the transaction, you specify which collections the transaction will write to.
 
-Once you create a transaction, you will be given a transaction reference number.
+Once you create a transaction, you will be given a transaction identifier.
 
 ### 2. Make API Calls
 
-Make API calls that include the transaction reference number to include them in the transaction. All actions will run, but they are not final until you commit the transaction.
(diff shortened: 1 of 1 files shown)


docs/transactions/index.md
```diff
@@ -1,4 +1,45 @@
...
+
+## How They Work
+
+Simply put, transactions are a way to aggregate, or group, multiple API calls. The process is as follows:
+
+### 1. Create a Transaction
+
+When you create the transaction, you specify which collections the transaction will write to.
+
+Once you create a transaction, you will be given a transaction reference number.
```
luisleite-macrometa: Maybe "transaction identifier" would be more appropriate than "transaction reference number", as the fact that the IDs are currently numbers is just an implementation detail subject to change in the future.

docs/transactions/index.md
```diff
@@ -1,4 +1,45 @@
...
+
+### 2. Make API Calls
+
+Make API calls that include the transaction reference number to include them in the transaction. All actions will run, but they are not final until you commit the transaction.
+
+### 3. Commit or Abort Transaction
+
+- To finalize all transaction API calls, commit the transaction.
```
luisleite-macrometa: What about "To make all changes made within the transaction permanent, commit the transaction."?

docs/transactions/index.md
```diff
@@ -1,4 +1,45 @@
...
+
+### 1. Create a Transaction
+
+When you create the transaction, you specify which collections the transaction will write to.
+
+Once you create a transaction, you will be given a transaction reference number.
+
+### 2. Make API Calls
+
+Make API calls that include the transaction reference number to include them in the transaction. All actions will run, but they are not final until you commit the transaction.
```
luisleite-macrometa: Again, "transaction identifier" may be a better option.

docs/transactions/index.md
```diff
@@ -1,4 +1,45 @@
...
+
+- To finalize all transaction API calls, commit the transaction.
+- To roll back the actions, perhaps because one of them failed, abort the transaction.
+
+## Supported Operations
+
+Supported operations include:
+
+- Read and write documents
+- Get the number of documents of collections
```
luisleite-macrometa: What about "Get the number of documents in collections"?

docs/transactions/index.md
```diff
@@ -1,4 +1,45 @@
...
+
+- To make all changes made within the transaction permanent, commit the transaction.
+- To roll back the actions, perhaps because one of them failed, abort the transaction.
+
+## Supported Operations
+
+Supported operations include:
+
+- Read and write documents.
+- Get the number of documents in a collection.
```
luisleite-macrometa, line 32: To keep this item consistent with the other sentences in this list, shouldn't we use "in collections" rather than "in a collection"?

luisleite-macrometa (approved):

docs/transactions/index.md
diana-macrometa: Where else would they read or write documents to? Are there options besides reading docs in a collection or writing them to a collection?

With the second bullet, "in a collection" is there because it is necessary. Although now I realize you could probably get the number of items in a particular KV Group, too.

luisleite-macrometa: Documents are only read from and written to collections.

Actually, it is possible to retrieve a list of KV groups in a KV collection within a transaction.

Since in the other items we have "truncate collections" rather than "truncate a collection", and "run queries and query workers" rather than "run a query or a query worker", I was wondering if we shouldn't write it as "get the number of documents in collections" rather than "get the number of documents in a collection". But maybe changing from "in a collection" to "in collections" has any subtle change in the meaning of the sentence that I, as a non-native English speaker, can't get. But it was just a guess.

diana-macrometa: Yes, if I said, "Get the number of documents in collections," that suggests that a user could get the number of documents in all collections, but you can only get the count for one at a time. That's why I changed it to singular.

diana-macrometa: Updated. How's this?

docs/transactions/index.md
```diff
@@ -28,8 +28,9 @@ Make API calls that include the transaction identifier to include them in the tr
 
 Supported operations include:
 
-- Read and write documents.
+- Read and write documents to collections.
 - Get the number of documents in a collection.
+- Get the number of key-value pairs in a collection.
```
luisleite-macrometa, line 33: I think this item is unnecessary, as it duplicates the previous item but for a KV collection. Maybe "Get and update the groups of a KV collection" or something similar could address the KV Group feature.

docs/transactions/index.md
```diff
@@ -1,4 +1,46 @@
...
+
+## Supported Operations
+
+Supported operations include:
+
+- Read and write documents to collections.
+- Get the number of documents in a collection.
+- Get the number of key-value pairs in a collection.
+- Truncate collections.
+- Run queries or query workers.
```
luisleite-macrometa: BTW I just recalled that it is a known bug that transactions are not working with query workers yet, and it likely will not get fixed on time for 0.18.0.

docs/transactions/index.md
```diff
@@ -1,4 +1,50 @@
...
+Transactions in Macrometa within a region are atomic, consistent, isolated, and durable (ACID).
+
+These ACID properties provide the following guarantees:
+
+- The _atomicity_ property makes transactions either complete in their entirety or have no effect at all.
+- The _consistency_ property ensures that no constraints or other invariants will be violated during or after any transaction.
+- The _isolation_ property hides the modifications of a transaction from other transactions until the transaction commits.
+- The _durability_ property ensures that operations from transactions that have committed will be made persistent.
+
+:::note
```
diana-macrometa, line 48: @maheshw2 - I would specifically like you to review this note.

maheshw2: Looks good to me

maheshw2 (approved):

PR #1053, opened 2023-11-30
18-00 - Transaction limits and limitations

diana-macrometa:
This topic adds information about limits and limitations for transactions.


docs/transactions/limits-limitations.md
```diff
@@ -1 +1,51 @@
...
+
+## Transactional Operation Limitations
+
+Transactional operation limitations within Macrometa are put in place to ensure the integrity and stability of the database system. These limitations cover the range of operations that are not supported within transactions, constraints on transaction implementation, and requirements for declaring collections for modification.
+
+### Unsupported Operations
+
+Certain actions are prohibited within transactions to maintain system integrity and performance. These include:
+
+- Creation and deletion of databases
```
luisleite-macrometa: Databases are usually referred to as (Geo) Fabrics in GDN.

docs/transactions/limits-limitations.md
```diff
@@ -1 +1,51 @@
...
+- Creation and deletion of indexes
+- Nesting a transaction within another transaction
+
+### Transaction Implementation Constraints
+
+Transactions are not optimized for long-running or voluminous tasks. They require all transactional information, such as record pointers and rollback data, to fit within the main memory. Additionally, transactions cannot be nested, and all collections that will be modified must be declared at the start of the transaction.
+
+## Atomicity in Multi-Collection Transactions
+
+Macrometa's approach to transaction atomicity is tailored for speed and efficiency but can be affected by server outages. In multi-collection transactions, atomicity across servers cannot be guaranteed if a server outage occurs during the commit phase.
```
luisleite-macrometa: This only applies to multi-node configurations, and multi-document transactions in a multi-node region are also subject to this limitation.

docs/transactions/limits-limitations.md
```diff
@@ -1 +1,51 @@
...
+
+## Handling Large Transactions
+
+Macrometa stores transaction data in RAM. Transactions that become too large are automatically committed and divided into smaller transactions. This approach helps manage memory usage but may affect the ACID properties of the original large transaction.
```
luisleite-macrometa: Intermediate commits are not enabled for user-defined transactions. If an operation causes the transaction size to exceed the provided limit (128MB by default), the operation fails.

docs/transactions/limits-limitations.md
diana-macrometa: So, what do I need to change here?

luisleite-macrometa: Something like:
"Macrometa's approach to transaction atomicity is tailored for speed and efficiency. As a consequence, on multi-node regions, the atomicity of multi-document/multi-collection transactions across servers cannot be guaranteed if a server outage occurs during the commit phase."

BTW we will have public documentation about multi-node support in GDN, right?

diana-macrometa: No, not for this release. We are not talking about multi-node support.

diana-macrometa: So I was just going to talk about multi-collection.

luisleite-macrometa: So multi-node is supposed to be transparent to the user, right? Unfortunately this is not always the case. Some of the limitations of transactions only apply to multi-node. Should we document these limitations as always present, although they only really happen on multi-node regions? Maybe Justin or Durga should be aware of this situation.

diana-macrometa: @elof and @durgagokina, we could use some guidance here.

diana-macrometa: @elof and @durgagokina , please opine.

@luisleite-macrometa , if they do not respond, then let's just remove everything that only applies to multinode. We'll have to add it back in when multinode goes GA.

elof: @diana-macrometa @luisleite-macrometa Since we aren’t likely to have a multi-node self-service platform in the near, or maybe even long term, and any dedicated multi-node deployment will definitely be an enterprise conversation where we will discuss requirements during a PoC I think it's ok to remove everything that only applies to multinode. @durgagokina let us know if you would prefer we keep the limitations in the docs

One other thing, I think it makes sense to list transactions between Query Workers and Streams, since they are very related to queries @diana-macrometa @luisleite-macrometa Since we aren’t likely to have a multi-node self-service platform in the near, or maybe even long term, and any dedicated multi-node deployment will definitely be an enterprise conversation where we will discuss requirements during a PoC I think it's ok to remove everything that only applies to multinode. @durgagokina let us know if you would prefer we keep the limitations in the docs

One other thing, I think it makes sense to list transactions between Query Workers and Streams, since they are very related to queries https://capture.dropbox.com/lIhhu08oJWh60zUh

diana-macrometa: Not sure what you mean. Could you post what you're talking about as a suggestion where you want it?

diana-macrometa: @luisleite-macrometa and @elof, what do we need to do to ship this? Multinode is not going with 18.0.

diana-macrometa:
@SpringNgMacrometa, this PR is well past SLA. Please help unblock.

docs/transactions/limits-limitations.md
luisleite-macrometa: @diana-macrometa, if multi-node is not going with 0.18.0, then you can remove section "Atomicity in Multi-Collection Transactions". Besides, section "Handling Large Transactions" must be removed since it is not correct, as explained in the comment.

diana-macrometa: Both sections are gone. Are we good?

luisleite-macrometa (approved):

PR #1054, opened 2023-11-30
18.00 - Transaction Attributes

diana-macrometa:
This PR adds information about transaction attributes to the Transactions section.


docs/transactions/transaction-attributes.md
```diff
@@ -0,0 +1,36 @@
...
+
+## `collections` Attribute
+
+The `collections` attribute is central to transaction management. It specifies which collections the transaction interacts with and in what manner:
+
+- `read`: Lists collections accessed only for reading. This is optional if `allowImplicit` is set to `true`.
+- `write`: Includes collections where data might be read or modified.
+- `exclusive`: Specifies collections for exclusive write access, preventing concurrent write operations. It essentially "locks" the collection.
+
+This attribute ensures data integrity and governs access levels to collections within a transaction. However, it can lead to deadlocks.
```
luisleite-macrometa: Actually, the deadlocks can only occur if a collection that is read from within the transaction is not declared in the `read` sub-attribute. So this is one more reason why declaring collections in `read` is recommended.

docs/transactions/transaction-attributes.md
```diff
@@ -0,0 +1,36 @@
...
+
+## `lockTimeout` Attribute
+
+`lockTimeout` sets a time limit in seconds for how long a transaction waits to acquire a lock on a collection. It's particularly relevant for transactions requiring exclusive access. A well-set `lockTimeout` can prevent transactions from stalling indefinitely, especially when set to 0, which removes waiting for locks.
```
luisleite-macrometa: In fact, a `lockTimeout` set to 0 causes the transaction to wait indefinitely on locks.
BTW the default lock timeout is 900 seconds.

docs/transactions/transaction-attributes.md
```diff
@@ -0,0 +1,36 @@
...
+
+## `maxTransactionSize` Attribute
+
+This attribute defines the maximum size of a transaction in bytes, capping it to ensure system stability. The limit is set to a maximum of 128 MB. Adjusting `maxTransactionSize` is crucial for maintaining performance and avoiding oversized transactions that could strain the system.
+
+## `waitForSync` Attribute
+
+`waitForSync` influences how transaction data is written to disk. When true, it ensures data is immediately synchronized to disk, offering higher data safety at the cost of performance. Conversely, false allows for delayed disk synchronization, improving performance but with a slight risk of data loss in the event of a crash.
```
luisleite-macrometa, line 30: Shouldn't `true` and `false` be formatted as code, like in the "`allowImplicit` Attribute" section?

diana-macrometa: If I was specifically talking about the entry `true`, then yes. However, I'm using "true" and "false" more generically.

That is true.
That is set to `true`.

It's kind of a weird gray area. I try to avoid overusing formatting, because if you have too much, the reader becomes blind to it.

docs/transactions/transaction-attributes.md
```diff
@@ -0,0 +1,36 @@
...
+
+- `read`: Lists collections accessed only for reading, which is recommended to prevent deadlocks. This is optional if `allowImplicit` is set to `true`.
+- `write`: Includes collections where data might be read or modified.
+- `exclusive`: Specifies collections for exclusive write access, preventing concurrent write operations. It essentially "locks" the collection.
+
+This attribute ensures data integrity and governs access levels to collections within a transaction. However, it can lead to deadlocks if a collection that is read from within the transaction is not declared in the read sub-attribute.
+
+## `lockTimeout` Attribute
+
+`lockTimeout` sets a time limit in seconds for how long a transaction waits to acquire a lock on a collection. It's particularly relevant for transactions requiring exclusive access. A well-set `lockTimeout` can prevent transactions from stalling indefinitely, especially when set to 0, causes the transaction to wait indefinitely. Default lock timeout is 900 seconds.
```
luisleite-macrometa: "A well-set `lockTimeout` can prevent transactions from stalling indefinitely, especially when set to 0, causes the transaction to wait indefinitely."

Isn't this sentence contradictory? First it suggests that setting a reasonable `lockTimeout` value can avoid transactions from getting stuck indefinitely. Then it says that setting `lockTimeout` to 0 causes the transaction to wait indefinitely, which is contrary to the idea of preventing transactions from stalling.

diana-macrometa: Good point. Updated.

docs/transactions/transaction-attributes.md
```diff
@@ -0,0 +1,36 @@
...
+
+## `collections` Attribute
+
+The `collections` attribute is central to transaction management. It specifies which collections the transaction interacts with and in what manner:
+
+- `read`: Lists collections accessed only for reading, which is recommended to prevent deadlocks. This is optional if `allowImplicit` is set to `true`.
+- `write`: Includes collections where data might be read or modified.
+- `exclusive`: Specifies collections for exclusive write access, preventing concurrent write operations. It essentially "locks" the collection.
+
+This attribute ensures data integrity and governs access levels to collections within a transaction. However, it can lead to deadlocks if a collection that is read from within the transaction is not declared in the read sub-attribute.
```
luisleite-macrometa: "it can lead to deadlocks if a collection that is read from within the transaction is not declared in the read sub-attribute."

Maybe it makes more sense to put this sentence in the description of  the `read` attribute.

luisleite-macrometa (approved):

PR #1068, opened 2024-01-04
Prerender settings

diana-macrometa:
This PR adds documentation about prerendering settings.


docs/photoniq/prerendering/prerendering-settings.md
```diff
@@ -0,0 +1,97 @@
...
+4. Scroll down to the setting that you want to view
+
+</TabItem>
+<TabItem value="api" label="REST API">
+
+Use our interactive API Reference with code generation in 18 programming languages to:
+
+- [Get prerender settings](https://www.macrometa.com/docs/apiPrerendering#/paths/api-prerender-v1-origins-origin--settings/get).
+- [Get all origin settings](https://www.macrometa.com/docs/apiPrerendering#/paths/api-prerender-v1-origins/get)
+- [Get settings for an origin](https://www.macrometa.com/docs/apiPrerendering#/paths/api-prerender-v1-origins/get)
```
sandro-macrometa: Here, you are referencing the wrong endpoint. This is the correct one:
`https://www.macrometa.com/docs/apiPrerendering#/paths/api-prerender-v1-origins-origin/get`
The label could be `Get origin settings` instead, as we have `Update origin settings` right below.

docs/photoniq/prerendering/prerendering-settings.md
```diff
@@ -0,0 +1,97 @@
...
+## Update Settings
+
+Use our interactive API Reference with code generation in 18 programming languages to:
+
+- [Update prerender settings](https://www.macrometa.com/docs/apiPrerendering#/paths/api-prerender-v1-origins-origin--settings/patch).
+- [Update origin settings](https://www.macrometa.com/docs/apiPrerendering#/paths/api-prerender-v1-origins-origin/patch)
+
+## Dynamic Prerendering Settings
+
+Dynamic Prerendering provides the following settings. Field names in the web console are different than those in the API responses. Refer to the API endpoint documentation for a full explanation of the API response.
```
sandro-macrometa: This is the same suggestion I left in the origins PR.
```suggestion
Dynamic Prerendering provides the following settings. Field names in the web console are different than those in the API responses. Refer to the API endpoint documentation for a full explanation.
```

sandro-macrometa (changes requested):

docs/photoniq/prerendering/prerendering-settings.md
diana-macrometa: Yes, I took some liberties with the labels, because this difference is really subtle and not easily understandable without the context of the API reference (and maybe not even with):

- Get origins
- Get origin

diana-macrometa: Updated

sandro-macrometa (approved):

PR #1077, opened 2024-01-10
18.00 - Functions update

diana-macrometa:
This PR documents changes made in [GUI-1821](https://macrometa.atlassian.net/browse/GUI-1821).

Users can now test EdgeWorker functions, create new versions, and change versions.

[GUI-1821]: https://macrometa.atlassian.net/browse/GUI-1821?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/akamai/edit-edgeworkers.md
```diff
@@ -0,0 +1,35 @@
...
+   A .tar file containing the following files is downloaded to your computer:
+   - service.js
+   - main.js
+   - config.js
+   - client.js
+   - bundle.json
+
+6. Unpack the .tar files with a tool or a command such as `tar -xvf file.tar`.
+7. Edit the files.
+   - Most of the functional code is contained in main.js.
```
dwivedianurag, line 40: most of the functional code is not in main.js , it is present in service.js and then corresponding code in client.js,config.js and main.js

diana-macrometa: Should I just take out the entire bullet point? Or do we want to try to clarify? And does it matter to the users?

dwivedianurag (approved):
added a small comment apart from that everything else looks good to me

docs/akamai/edit-edgeworkers.md
dwivedianurag: it might matter to the user so if we are keeping it   we can go in detail what every file contains ... or else completely remove this bullet

diana-macrometa: Could you give me descriptions of all the files to include in the docs?

diana-macrometa:
@dwivedianurag, please give me the descriptions so we can wrap this up.

docs/akamai/edit-edgeworkers.md
dwivedianurag: client.js ---> In this we define the endpoints and logic  to execute the query worker and get the next batch in case of edgeworker via query worker, in case of stream publisher it is used for posting the message to the configured stream .
In case of edgeworker for stream query it  is used for posting queries to the stream worker code.

config.js --- for all type of supported edgeworkers it is here where we define the different variables to be used in the edgeworker . C8_URL -- being the gdn url . Fabric being the fabric you want to connect to , MACROMETA_ORIGIN_NAME is used for PM user variable and is used by the property to ensure  the request gets routed to the right macrometa origin and is passed as a header in the apicalls made from the edgeworker
in case of edge worker via queryworker  this file as QUERY_WORKER_NAME which is self explanatory 
in case of edge worker via stream publisher this has STREAM_APP_NAME which is your stream worker name and STREAM_NAME which is the name of the stream
incase of edgeworker via stream query it has STREAM_APP_NAME which is your stream worker name

service.js --> it is here we define end to end logic for what would happen when the api call is made and what should be the logic before and after the apicalls



main.js ---> it is the main entry of the code here you define your edgeworkers event handler to process incoming request to and outgoing response

diana-macrometa: Great stuff. How does my reformatted list look?

diana-macrometa:
@dwivedianurag, is this ready for approval?

PR #1103, opened 2024-02-07
Pull out priority queue in seperate section and give more information…

joubert-macrometa:
… about how it works.


docs/photoniq/vwrs/configure-vwrs-edgeworker.md
```diff
@@ -62,85 +62,6 @@ A request configuration is an optional object that contains configuration parame
...
-```bash
-GET /checkout?waiting-room-priority=20
-```
-
-The example adds the request to the waiting room with a priority of 20.
-
-```js
-export async function onClientRequest(request: EW.IngressClientRequest) {
-  const queryParams = new URLSearchParams(request.query);
-  const reqPriority = parseInt(queryParams.get("waiting-room-priority"), 10);
```
manoj-desilva, line 105: shouldn't this value be 20?
there are few other places that have 10 in this PR.

naimishvora-macrometa: @manoj-desilva that is redix means parseInt function will return explicitly specify base-10 integer so it is okay.

naimishvora-macrometa (approved):
LGTM

docs/photoniq/vwrs/configure-vwrs-edgeworker.md
manoj-desilva: Thanks @naimishvora-macrometa for clarification.

manoj-desilva (approved):

diana-macrometa (commented):
Why did you put this content in a separate section? Isn't this needed for configuring the EdgeWorker? 

I would prefer all the content for configuring the EdgeWorker be kept in one topic.

docs/photoniq/vwrs/configure-vwrs-priority.md
```diff
@@ -0,0 +1,85 @@
+---
+sidebar_position: 30
+title: Configure a VWRs Priority Waiting Room
+---
+
+Priority waiting rooms allow you to prioritize traffic based on the priority of the request. Requests are assigned a priority level and then dequeued at different rates based on the priority level. This feature helps handle various types of traffic, such as high-priority, regular, and low-priority.
```
diana-macrometa, line 6: If we have a separate topic for priority waiting rooms, then we need separate topics for the other waiting roo types to be consistent.

docs/photoniq/vwrs/configure-vwrs-priority.md
```diff
@@ -0,0 +1,85 @@
...
+
+VWRs dequeues requests by assigning weights to different priority levels descendingly, with the weight halving for each subsequent level. For instance, if there are three priority levels, the highest priority receives a weight of 4, the next lower priority gets a weight of 2, and the lowest priority is assigned a weight of 1. This creates a total weight summing up to 7 for all priorities combined. A formula determines the distribution of dequeued items from each priority level: RateLimit \* (Weight of Specific Priority Level / Total Weight of All Priorities). If a priority level has no requests to dequeue, its weight is redistributed among the remaining priorities, recalculating the weights based on the active queues. Additionally, if there are any remaining items after distribution, those are allocated to the lowest priority level.
```
diana-macrometa, line 8: This is great content. If you wanted to get more into priorities, we could write a concept topic similar to [VWRs Queue Types](https://www.macrometa.com/docs/photoniq/vwrs/queue-types).

diana-macrometa (changes requested):

docs/photoniq/vwrs/configure-vwrs-priority.md
joubert-macrometa: Priority queues are an optional configuration.  My guess would be 80+% will never use priority queues.  So thought we would pull it out of getting started because you don't need it to getting started.

Priority queues are not a different queue type.  It is configured on top of the queue type.  So a FIFO queue can have priority capabilities configured and can any of the other queue types.  You can never configure a priority queue by itself.  It needs to be part of the queue type.

joubert-macrometa:
Not going to use

PR #1136, opened 2024-04-02
Update P3 index.md

diana-macrometa:
This PR adds a feature list to the P3 Landing page.

### DIFF
--- a/docs/photoniq/p3/index.md
+++ b/docs/photoniq/p3/index.md
@@ -32,7 +32,7 @@ Key capabilities of P3 include:
 
 - **Dynamic Content Updating and Caching**: Through offline analysis and real-time optimizations, P3 dynamically updates and caches content to achieve peak performance.
 
-- **Advanced Compression Techniques**: Applies cutting-edge image and content compression strategies to minimize load times and enhance page responsiveness.
+- **Advanced Compression Techniques**: Applies cutting-edge content compression strategies to minimize load times and enhance page responsiveness.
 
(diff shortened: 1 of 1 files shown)


db-macrometa (changes requested):
So far we don't have anything pertaining to **image compression** or any image optimizations for that matter under Advanced Compression Techniques

diana-macrometa:
> So far we don't have anything pertaining to **image compression** or any image optimizations for that matter under Advanced Compression Techniques

If you give me something for the list, I am happy to add it. This is the list I was given, and it seems to match what is in the UI.

Keep in mind, if we have things in the docs, then we will need to tell the user how they can enable them (API, UI, or talk to MM).

db-macrometa:
I don't think we have any UI or API that would suggest that we support image compression or optimization @diana-macrometa Unless I missed it, you can safely remove this specific wording, i.e image compression and just leave content compression.

diana-macrometa:
> I don't think we have any UI or API that would suggest that we support image compression or optimization @diana-macrometa Unless I missed it, you can safely remove this specific wording, i.e image compression and just leave content compression.

AH, I misunderstood what you were saying. This is why I suggest that editors use the Git tools to specifically comment and suggest on the line where they want changes, because that helps prevent misunderstanding.
image

diana-macrometa:
All right, I updated what I think you were referring to. Anything else?

edgargarciamacrometa (commented):
👍

db-macrometa (approved):

PR #1155, opened 2024-04-17
P3 on demand validations

diana-macrometa:
This PR adds on-demand validation documentation.


docs/photoniq/p3/manage-on-demand-validations.md
```diff
@@ -0,0 +1,48 @@
+---
+sidebar_position: 45
+title: Manage On-demand Validations
+---
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+After you create a policy and PhotonIQ Performance Proxy (P3) gets a request for an optimized page, P3 validates the page by running visual and functional tests. You can perform additional validations on-demand in the PhotonIQ Performance Proxy dashboard.
```
dwivedianurag: we can also word something like this that if you want to validate a specific url which comes under the url pattern for which you have configured you can use this feature. There will also be a limit added to it for how many times in an hour you can run this

dwivedianurag (approved):
ok

PR #1161, opened 2024-04-23
Prerender htmlSelector

diana-macrometa:
This page provides instructions and considerations for finding the htmlSelector string.

### DIFF
--- a/docs/photoniq/prerendering/synthetic-interactions/find-htmlselector.md
+++ b/docs/photoniq/prerendering/synthetic-interactions/find-htmlselector.md
@@ -59,7 +59,9 @@ If the element is within a Shadow DOM:
 
 #### Consider Specificity and Efficiency
 
-The automatically generated selector might not always be the most efficient or simplest. It might be overly specific. Simplify the selector if possible, keeping only the parts that are necessary to uniquely identify the element. For example, if the generated selector is `html > body > div:nth-child(3) > div > p > a`, and you know the `a` element is unique within its parent `p`, you might simplify this to `p > a`.
+The automatically generated selector might not always be the most efficient or simplest. It might be overly specific. Simplify the selector if possible, keeping only the parts that are necessary to uniquely identify the element.
+
+For example, if the generated selector is `html > body > div:nth-child(3) > div > p > a`, and you know the `div` element has the id `login-container`, you might simplify this to `#login-container > p > a`.
 
 ## Testing the `htmlSelector`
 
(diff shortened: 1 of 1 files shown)


docs/photoniq/prerendering/synthetic-interactions/find-htmlselector.md
```diff
@@ -0,0 +1,140 @@
...
+#### For Shadow DOM Elements
+
+If the element is within a Shadow DOM:
+
+- You might need to use JavaScript to access the shadowRoot of the custom element. For example: `document.querySelector('custom-element').shadowRoot.querySelector('element')`.
+- Some browsers' Developer Tools allow you to inspect elements within Shadow DOM directly. Ensure that the option to show user agent shadow DOM is enabled in the settings of the Developer Tools.
+
+#### Consider Specificity and Efficiency
+
+The automatically generated selector might not always be the most efficient or simplest. It might be overly specific. Simplify the selector if possible, keeping only the parts that are necessary to uniquely identify the element. For example, if the generated selector is `html > body > div:nth-child(3) > div > p > a`, and you know the `a` element is unique within its parent `p`, you might simplify this to `p > a`.
```
sandro-macrometa: For me, this is not a good example because `p > a` means the first link inside all paragraphs on the HTML. Here, a lot of elements can be selected, as it will search for all paragraphs in the HTML. Here, maybe we should give an example that uses an ID because `html > body > div:nth-child(3) > div > p > a` is already more specific and efficient than `p > a`.

I suggest something like this:

For example, if the generated selector is `html > body > div:nth-child(3) > div > p > a`, and you know the `a` element has the id `login-button`, you might simplify this to `#login-button`.

Or this one, for not being too obvious:

For example, if the generated selector is `html > body > div:nth-child(3) > div > p > a`, and you know the `div` element has the id `login-container`, you might simplify this to `#login-container > p > a`.

sandro-macrometa (changes requested):
Awesome work here, Diana. I just left one suggestion.
Let me know what you think.

sandro-macrometa:
@diana-macrometa, one thing more we need to add. For the shadow dom we need to prefix the selector with `>>>`.
This is mentioned in the notion document here: 
https://www.notion.so/macrometa/How-to-Find-htmlSelector-for-Synthetic-Interactions-08cbb4e9e4574080b99a5c90660a0714?pvs=4#98158dfb46ee43c4a53b2a82e1471ead 
Also here:
https://www.notion.so/macrometa/How-to-Find-htmlSelector-for-Synthetic-Interactions-08cbb4e9e4574080b99a5c90660a0714?pvs=4#5a6b699db2ca4ab5b4752b949e48ea75

diana-macrometa:
Good catch! Thanks, @sandro-macrometa. Any other edits?

sandro-macrometa (approved):

sandro-macrometa:
> Good catch! Thanks, @sandro-macrometa. Any other edits?

Looks good to me!

PR #1162, opened 2024-04-24
Update prefetching tab docs

diana-macrometa:
This PR updates documentation for Prefetching tab tasks.

**Reviewers** - A couple questions in comments.


docs/photoniq/prerendering/manage-prefetching/configure-prefetching.md
```diff
@@ -0,0 +1,99 @@
...
+
+</TabItem>
+<TabItem value="api" label="REST API">
+
+Use our interactive API Reference with code generation in 18 programming languages to [get prefetching configuration](https://www.macrometa.com/docs/apiPrerendering#/paths/api-prerender-v1-origins-origin--prefetch-configs/get).
+
+</TabItem>
+</Tabs>
+
+## Delete Prefetching Configurations
```
diana-macrometa, line 97: Why can't we delete configs in the web console?

docs/photoniq/prerendering/manage-prefetching/configure-prefetching.md
```diff
@@ -0,0 +1,99 @@
...
+- **Scheduling** - Cron expression entered to trigger the prefetching job.
+- **Devices** - Icon indicating whether the prefetch was rendered for just desktop or desktop and mobile.
+- **Actions** - Icon that allows you to update.
+
+## Create Prefetching Configurations
+
+<Tabs groupId="operating-systems2">
+<TabItem value="console" label="Web Console">
+
+A new prefetch configuration appears in the web console when you add a new origin in the Origin Settings tab. For more information about viewing and adding origins, refer to [Manage Origins](../manage-origins/).
```
diana-macrometa, line 50: Do we need to say something about the sitemap? As I recall, the sitemap is needed if you are doing prefetching rather than on-demand.

docs/photoniq/prerendering/manage-prefetching/configure-prefetching.md
sandro-macrometa: This will be available in `1.2.1` I think. The next release is `1.2.0`.

docs/photoniq/prerendering/manage-prefetching/configure-prefetching.md
sandro-macrometa: The sitemaps column just reminds the user that the prefetching will happen on all URLs of the sitemap.
Those sitemaps are only used when the configuration for the origin is active.

sandro-macrometa (approved):
LGTM

docs/photoniq/prerendering/manage-prefetching/configure-prefetching.md
sandro-macrometa: In that release, I think we're going to change that screen's behavior. Instead of displaying all origins, even without a prefetch config, we are going to display only the origins with a prefetch config. Then, you will be able to add new configurations by clicking on a new button outside the table. 

We'll let you know so you can also update the docs.

PR #1167, opened 2024-04-29
Add doc for support of native apps.

joubert-macrometa:
We have never documented how a native app (iOS, Android, etc) can used with the waiting room, because we were missing a usable API. This PR documents the new "position" API and how to use it.

### DIFF
--- a/static/openapi/vwrs-spec.json
+++ b/static/openapi/vwrs-spec.json
@@ -1,2258 +1,2428 @@
 {
-  "openapi": "3.0.0",
-  "info": {
-    "title": "PhotonIQ Virtual Waiting Rooms API",
-    "description": "API documentation for the PhotonIQ Virtual Waiting Rooms",
-    "version": "1.26.0"
-  },
-  "servers": [
-    {
-      "url": "https://api-demo-vwr.paas.macrometa.io",
-      "description": "PhotonIQ Virtual Waiting Rooms API"
-    }
-  ],
-  "components": {
-    "securitySchemes": {
-      "APIKeyAuth": {
-        "type": "apiKey",
-        "in": "header",
-        "name": "Authorization"
-      }
+    "openapi": "3.0.0",
+    "info": {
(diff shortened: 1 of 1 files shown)


diana-macrometa:
@joubert-macrometa , please add a description explaining what changed. I see the API spec changed a great deal.

For the native app file, what is the purpose of the file? Is it much different than the getting started content? It seems a bit redundant, so I want to understand how it adds value to the documentation.

joubert-macrometa:
@diana-macrometa I am getting @todor-macrometa to review the api spec.  It should be a few lines added only, so don't understand why these massive changes.

joubert-macrometa:
We have never documented how a native app (iOS, Android, etc) can used with the waiting room, because we were missing a usable API.  This PR documents the new "position" API and how to use it.

diana-macrometa:
In that case, let's have a page explaining the position API, what it is and why to use it. 

Right now, the topic really only mentions position in passing. The rest of it talks about "native apps," and I do not understand how this is different than any of the rest of our documentation. I think we could potentially work some of it, especially the explanation and pseudocode, into the [Get Started with VWRs](https://www.macrometa.com/docs/photoniq/vwrs/get-started-vwrs) topic, which right now is pretty sparse. 

The best practices could also be their own topic, VWRs Best Practices, as they seem pretty generic.

joubert-macrometa:
Supporting native apps is a special case.  It is a replacement for the web browser.

The getting started guide seems to be how you configure the edgeworker and the vwrs.

But Native apps have nothing to do with that.  That is why I separated it.  It is used by external clients who interact with the waiting room that are not a web browser.

The API is documented in vwrs-sepc.json.  But we need to fix it because it has all these changes that should not be there.  Hopefully, that will be fixed in the next 24 hours.  

We might need to go to voice for me to explain it better.

joubert-macrometa:
Not using since it is being done in a different PR.

PR #1169, opened 2024-04-30
Prerender update policies

diana-macrometa:
Added information about how global and specific policies interact.


sandro-macrometa (approved):
Nice, this is what I mentioned in the previous PR. Scrolling behaves differently.

PR #1176, opened 2024-05-06
Update p3

diana-macrometa:
P3 UI updates.


docs/photoniq/p3/manage-p3-policies.md
```diff
@@ -55,6 +48,15 @@ Perform the following steps to create a new P3 policy:
       or provide regex:
         - ^/us/category/.*$
         - ^/us/product/.*$
+   - **Sample Path** - NEED DEFINITION
```
diana-macrometa: @dwivedianurag - @nobuyoshi-macrometa explained that this is a URL that needs to match the regex pattern, but I do not understand why as a user I am entering this as well as the pattern.

dwivedianurag: you are entering the pattern cause you want to group the pages and in here you are entering an example of the path which matches those patterns. This will later be converted to let the user validate the regex. Currently this is required by TCP to purge the cache and get the optimizations before hand without any request coming to P3

dwivedianurag (approved):
ok please update the definition for sample path before proceeding

docs/photoniq/p3/manage-p3-policies.md
diana-macrometa: How's this?

docs/photoniq/p3/manage-p3-policies.md
```diff
@@ -55,6 +48,15 @@ Perform the following steps to create a new P3 policy:
       or provide regex:
         - ^/us/category/.*$
         - ^/us/product/.*$
+   - **Sample Path** - An example of a path that matches the URL pattern. This will later be used to validate the regex and for automated cache purging.
```
dwivedianurag: @diana-macrometa we should remove this This will later be used to validate the regex and for automated cache purging

dwivedianurag (approved):
please review the commit before merging

PR #1185, opened 2024-05-20
EDS-9: Add EDS Client Documentation

dlazarenko-macrometa:
Added documentation for `https://github.com/Macrometacorp/photoniq-eds-ws/` driver into PhotonIQ docs.
I added the module `photoniq-eds-ws.js` as a static asset into docs. It is temporary solution. In the future it will be added to a public package manager like `https://www.jsdelivr.com/`


felipe-lopes-macrometa (approved):

docs/photoniq/event-delivery/clients/index.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Basic Example
+
+Connect to Event Delivery Service, retrieve and subsribe to SQL:
+```js
+let config = {
+    host: "<YOUR-PHOTONIQ>.photoniq.macrometa.io",
+    customerId: "<YOUR-CUSTOMER-ID>",
+    apiKey: "<YOUR-API-KEY>",
+    fabric: "<YOUR-FABRIC>",
```
joubert-macrometa, line 29: While not really a doc comment...

Strange that you are exposing fabric here.
I would have thought that if you are using photoniq-auth, you would get the fabric that is tied to the apikey.

The same for customerId.  Does'nt the apikey identify the customerId?

docs/photoniq/event-delivery/clients/index.md
```diff
@@ -0,0 +1,61 @@
...
+};
+
+let connection = PhotoniqEdsWs.connect(config);
+
+let querySet = connection.querySet();
+
+querySet.retrieveAndSubscribe("SELECT * FROM <YOUR-COLLECTION> WHERE _key='<KEY-OF-DOCUMENT>'", (event) => {
```
joubert-macrometa: Another strage is why use _key?  Why not just use "key"?

docs/photoniq/event-delivery/clients/index.md
```diff
@@ -0,0 +1,61 @@
...
+})
+```
+> Note: The example uses the WebSocket client. To use an SSE connection, replace `PhotoniqEdsWs` with `PhotoniqEdsSse`.
+
+## Config Schema
+
+| **Property** | **Type** | **Requred** | **Description** |
+|----------------------|-----------|-----------|-----------------------------------|
+| `host` | `string` | Yes | Host of the connectionn   |
+| `fabric` | `string` | No | Fabric to be used. Default is `_system`   |
```
joubert-macrometa, line 51: You are exposing GDN stuff here.  I thought photoniq services hid all GDN stuff?  This probably goes to my above question why is this even being exposed.

docs/photoniq/event-delivery/clients/connection.md
```diff
@@ -0,0 +1,50 @@
...
+sidebar_position: 10
+title: Connection
+---
+
+It maintains a WebSocket/SSE connection:
+
+## Supported methods
+
+### connect
+
```
joubert-macrometa, line 11: What is the intent of this section?

You define methods.  But I don't know what to do with them.  For example, getConfig() returns a Config.
Is Config an opaque type?  Or is there information inside it that I need?

I don't understand, as a user of the Client SDK how can I use this information that is being provided here?

joubert-macrometa: OK, I found that config is defined on the main "Client SDKs" page.

Still, when I am looking at SDKs, I want to look at all the parameters.  This does not define the APIs well.

We are going to need to get help from the doc person to give their thoughts.

joubert-macrometa (changes requested):
I wanted to hear some of your thoughts on my comments.

docs/photoniq/event-delivery/clients/index.md
dlazarenko-macrometa: @joubert-macrometa - it is how now EDS works, `customerId` and `fabric` will be deleted when full functionality will be implemented in EDS. 
cc @manoj-macrometa

docs/photoniq/event-delivery/clients/index.md
dlazarenko-macrometa: @joubert-macrometa - the same here. `customerId` and `fabric` will be removed from the table. 
cc @manoj-macrometa

docs/photoniq/event-delivery/clients/index.md
dlazarenko-macrometa: @joubert-macrometa - I used a common property `_key` from collection. Should I change to `key`?

docs/photoniq/event-delivery/clients/connection.md
maryann-macrometa: I'm still familiarizing myself with the docs, but I agree. Some users may get frustrated. Let's make it easier for them to find everything they need in one place and link to what needs any extra definition. 

How about after defining the methods, we add any required parameters and show what calling that method returns? This helps let them know if they're on the right track.

dlazarenko-macrometa: @joubert-macrometa - it is not clear how to improve. Do you mean to provide link to `Config` Schema table for `getConfig()` method? Or even put all methods into a table?

joubert-macrometa: @dlazarenko-macrometa I am also not 100% sure.  Just what we have now is not great.

What @maryann-macrometa is saying is probably close to what we need.  Mimic what swagger is doing.  Knowing what the inputs are and what the outputs are in one place will make, me as a developer, much happier :-) 

@maryann-macrometa as you come up to speed on the docs, keep an eye if we have done something like this before.  Documenting the SDK.  Maybe we leverage that for now until you can help us come up with a more coherent strategy for this.

docs/photoniq/event-delivery/clients/index.md
joubert-macrometa: Understood.

docs/photoniq/event-delivery/clients/index.md
joubert-macrometa: I find the _key terrible.   My suggestion would be to remove the "_" and just call it key.

joubert-macrometa (approved):
@dlazarenko-macrometa I am approving.

I don't know how we can improve the api documentation.  

Maybe we leave it for now and @maryann-macrometa can help us clean this up in another PR.

docs/photoniq/event-delivery/clients/index.md
dlazarenko-macrometa: @joubert-macrometa - If I understood correctly it can be like this: `("SELECT * FROM <YOUR-COLLECTION> WHERE key=<YOUR-KEY>"`

docs/photoniq/event-delivery/clients/connection.md
maryann-macrometa: @joubert-macrometa I'll do that.

docs/photoniq/event-delivery/clients/index.md
joubert-macrometa: yup

dlazarenko-macrometa: @joubert-macrometa - I've updated the SQL.

docs/photoniq/event-delivery/clients/connection.md
dlazarenko-macrometa: @joubert-macrometa, @maryann-macrometa - I added link to `Config`

PR #1186, opened 2024-05-21
[VWROOM-422] Rename `domain` to `waitingroom`

todor-macrometa:
Jira: https://macrometa.atlassian.net/browse/VWROOM-422

Please don't merge this yet, as VWRS 2.0.0 is still WIP. I'll merge it as soon as 2.0.0 is out.

### DIFF
--- a/lib/openapi/revisions/vwrs-spec-2.0.0-20240521.json
+++ b/lib/openapi/revisions/vwrs-spec-2.0.0-20240521.json
@@ -3,11 +3,14 @@
   "info": {
     "title": "PhotonIQ Virtual Waiting Rooms API",
     "description": "API documentation for the PhotonIQ Virtual Waiting Rooms",
-    "version": "1.22.0"
+    "version": "2.0.0"
   },
   "servers": [
     {
-      "url": "https://api-demo-vwr.paas.macrometa.io"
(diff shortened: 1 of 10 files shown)


lib/openapi/revisions/vwrs-spec-2.0.0-20240521.json
```diff
@@ -3,11 +3,14 @@
   "info": {
```
todor-macrometa, line 3: This is a completely different file, not a rename of an existing one.
- I've deleted static/swagger/vwrs-spec.json, because it seems to be (1) out of date and (2) not used.
- I've added the vwrs-spec-2.0.0-20240521.json file, because the instructions page says that it should be added (https://www.notion.so/macrometa/Steps-to-update-PhotonIQ-VWRS-API-reference-documentation-e451f63422c64a1d9e29ee12b4f1c05c)

joubert-macrometa: I don't think you even need this.  Just delete it completely.  The documentation above should be changed to not create this anymore.  This goes back to the days we were storing everything in the swagger director.

joubert-macrometa (approved):

PR #1187, opened 2024-05-31
Editing faas

maryann-macrometa:
@dlozina @joubert-macrometa I had to manually copy and paste locally, hence the long time. Please check these new edits for errors, especially the code part. I broke it down and would like to know if it's still accurate and reads well. Some fields are still in progress; I'll work on them over the weekend. Thank you again for the help.


ghost (approved):
👍

PR #1189, opened 2024-06-05
edited the eds docs

maryann-macrometa:
Made some edits to EDS, especially the quickstart section. Want to be sure it's right .


docs/photoniq/faas/01-release-notes.md
```diff
@@ -3,5 +3,14 @@ sidebar_position: 2
 title: Release Notes
 ---
 
-## FaaS v1.001
-This represents the very first version of the PhotonIQ Function as a Service(FaaS)
\ No newline at end of file
+## May 2024 Release
+### v1.001
```
ghost: We use semver notation for vesions: https://semver.org/
This needs to be 1.0.0

static/img/photoniq/event-delivery/EDS-ss-showing-connected-with-data.png
ghost, line 1: It is better to use GUI images for docs, and when we need to display terminal output, we can use `bash` code comments. It is easier for developers to see what is happening, and they can also copy-paste from the docs, for example, the same filter while testing. What is marked on the image my suggestion is to move it to a line text explanation

ghost (changes requested):
Very good, minor comments are added.

docs/photoniq/faas/01-release-notes.md
maryann-macrometa: fixed

ghost (approved):
👍

PR #1229, opened 2024-08-07
Collections revamp

maryann-macrometa:
Restructured collections.


joubert-macrometa:
FYI, there are two photoniq images in this PR.

